PR #53, opened 2019-05-20
Implement simple maps support

Fabi2607:
The location data provided by the database can be used to create an
Map intent and show the summit in a user defined mapping
application (e.g. OsmAnd, Google Maps).

Signed-off-by: Fabian Kantereit <fabian.kantereit@tsaro.io>

### DIFF
--- a/app/src/main/java/com/example/paetz/yacguide/RouteActivity.java
+++ b/app/src/main/java/com/example/paetz/yacguide/RouteActivity.java
@@ -4,10 +4,12 @@ import android.app.Dialog;
 import android.content.Intent;
 import android.graphics.Color;
 import android.graphics.Typeface;
+import android.net.Uri;
 import android.os.Bundle;
 import android.view.View;
 import android.widget.LinearLayout;
 import android.widget.TextView;
+import android.widget.Toast;
 
 import com.example.paetz.yacguide.database.Ascend;
 import com.example.paetz.yacguide.database.Comment.RockComment;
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/example/paetz/yacguide/RouteActivity.java
```diff
@@ -50,6 +51,15 @@ public void back(View v) {
         finish();
     }
 
+    public void showMap(View v) {
+        Uri gmmIntentUri = Uri.parse("geo:"+ _rock.getLatitude() + "," + _rock.getLongitude());
```
paetz: could be final

app/src/main/java/com/example/paetz/yacguide/RouteActivity.java
```diff
@@ -50,6 +51,15 @@ public void back(View v) {
...
+        Intent mapIntent = new Intent(Intent.ACTION_VIEW, gmmIntentUri);
+        if (mapIntent.resolveActivity(getPackageManager()) != null) {
+            startActivity(mapIntent);
+        }
```
paetz: If there is no map tool available, you should show an according error message (using Toast) like "Keine Karten-App verfügbar."

app/src/main/java/com/example/paetz/yacguide/RouteActivity.java
```diff
@@ -50,6 +51,15 @@ public void back(View v) {
         finish();
     }
 
+    public void showMap(View v) {
+        Uri gmmIntentUri = Uri.parse("geo:"+ _rock.getLatitude() + "," + _rock.getLongitude());
+        Intent mapIntent = new Intent(Intent.ACTION_VIEW, gmmIntentUri);
+        if (mapIntent.resolveActivity(getPackageManager()) != null) {
+            startActivity(mapIntent);
```
paetz: I tested with google maps on the emulator and the coords were NOT represented with a marker. Can we somehow make sure that the map tool provides a proper marker for the given coords?

Fabi2607: Good question, we could request a whitelisted set of map tools...

Fabi2607:
I will address your comments, as soon as I find time. Just wanted to share my modifications in case someone finds them useful.

app/src/main/java/com/example/paetz/yacguide/RouteActivity.java
Fabi2607: I investigated a bit more... There is absolutely now way to tell the Uri that there should be a marker. We can either query different apps one by one, leave it as is or integrate a map toolkit (which I considered too much work right now)

paetz: Ok, then we leave it as is for now, since it's better than nothing

PR #57, opened 2019-07-13
osm map support

Fabi2607:
Due to the bad weather I kind of went crazy and got stuck in a long programming session.

I do not expect to get the stuff merged in the current state, but also don't want to keep what I am doing to myself.

What has changed:
- drawer style navigation, providing more usable space and a more "native" app feeling
- proper map support using mapforge (including clustering of symbols)

Things that are still a bit wonky:
- Map download is rough now, needs proper management for the download state, selection of areas etc.

Testing:
I have been running this development tree the whole time. YMMV.

Screenshot_20190713-152557
Screenshot_20190713-152612


Fabi2607:
For merging this: I will keep this draft open, move single changes to separate pull requests and rebase this branch. 

I will keep this PR open for general discussions

paetz:
Great work. I try to find some time at the weekend for reviewing but this will take some time. Beforehand:

1. I don't have a problem with using Kotlin, but if we do, we should transfer everything to Kotlin.
2. I don't see a reason why we should not merge your stuff :)
3. The map support is great but I also see problems:
3.a) I guess you will need to download a huge amount of data since you can only download big areas like the whole CZ right?
--> You would need to download certain tiles, depending on the locations of the rocks that are present in your local database.
--> Syncing of rocks would take much more time when downloading a map section additionally.
3.b) Where is the huge amount of map data stored? Intern phone memory may be too small.
3.c) Because of 3.a and 3.b, I would suggest to consider the opportunity of sending all rock locations of a sector to the external map app (if possible) and display it there. This also reduces complexity and maintenance work a lot.
4. Have you ensured that the code is still compatible with Android 5.1?

Fabi2607:
>  I try to find some time at the weekend for reviewing but this will take some time
Sure, I still know of quite a few rough edges, so don't spent to much time on the details. I also will not find much time to do much work on this over the next few weeks.

>     1. I don't have a problem with using Kotlin, but if we do, we should transfer everything to Kotlin.
Great. That was my biggest concern and should be easy to do.

>     2. I don't see a reason why we should not merge your stuff :)
Alright, just usually try to pack my changes in a more manageable way.

>     3. The map support is great but I also see problems:
>        3.a) I guess you will need to download a huge amount of data since you can only download big areas like the whole CZ right?
I would keep the map download separate and want to create a menu for selecting the map sources/enable online maps. There is no reason to keep all of the maps downloaded, even if you decide to download the whole database.

>        --> You would need to download certain tiles, depending on the locations of the rocks that are present in your local database.
>        --> Syncing of rocks would take much more time when downloading a map section additionally.
>        3.b) Where is the huge amount of map data stored? Intern phone memory may be too small.

The maps are in the mapforge binary format. Depending on the map source this is between 50-500 MB per area (i.e. Saxony, CZ etc.) Location is currently the data directory of the map (getExternalFilesDir())

>        3.c) Because of 3.a and 3.b, I would suggest to consider the opportunity of sending all rock locations of a sector to the external map app (if possible) and display it there. This also reduces complexity and maintenance work a lot.

There is no real defined protocol to do that, AFAIK. We would then depend entirely on a single app. Also we loose the opportunity to cluster while zooming out etc. 
 
>     4. Have you ensured that the code is still compatible with Android 5.1?
Hopefully it is, but I have only tested on 7.1 so far.

paetz:
Ok, I'd suggest the following steps to keep it clean (since we don't have unit tests):
1. PR: Add CI/CD to the project that builds and deploys a dev version executable for each commit to master. 
2. PR: Add .gitignore and remove unnecessary files
3. PR: Move all java files to Kotlin (without changing the app's behavior).
4. PR: UI redesign
5. PR: Add OSM map support
6. PR: Adapt CI/CD script to build and deploy releases from release branches. Create a release branch and let CI/CD deploy the first beta version.

Did I miss anything? We should create issues for each point.

Fabi2607:
That was exactly how I planned to do it. The first step of removing unnecessary files that are not build related is already done in #56 .

Fabi2607:
Kotlin conversion is done in #61 , current work branch has been rebased

chrgernoe:
> 1. PR: Add CI/CD to the project that builds and deploys a dev version executable for each commit to master.

I implemented that with the PR for #71.

chrgernoe:
> 4\. PR: UI redesign

Is somebody working at this PR? There are some other features or issues which depend on the new UI design, e.g. #72, #73 or #72.

Fabi2607:
> > 4. PR: UI redesign
> 
> Is somebody working at this PR? There are some other features or issues which depend on the new UI design, e.g. #70, #72, #73 or #74.

Yes, I have been working on this from time to time, and updating it with fixes and small features I wanted on short notice.

Fabi2607:
This merge request still has a few rough edges, but overall I feel like we should work on getting this merged, so we can work together more effectively.

I promise this is the last of these really big CRs and the next ones will be self contained changes. To this code.

The most critical thing is, that currently the download of the map data only has a button and no proper menu or setting screen (offline/online map source etc.)

Fabi2607:
@chrgernoe Can you tell me why the build keeps failing? It seems like OpenSSL errors out because it does not have an IV. I guess this is due to the branch being in a different repositories and Travis trying not to leak the environment variables?

chrgernoe:
I think this is because of #77. The current workaround is not commit to this branch and open an new one directly in this repository.

Fabi2607:
> @Fabi2607 I think this is because of #77. The current workaround is not commiting to this branch. Open an new one directly in this repository.

Ack, I will open a new branch and then close this PR one in favor of it.

Fabi2607:
Closed in favor of #81 #82 and #83

PR #71, opened 2019-10-22
Travis CI Integration

### LINKED ISSUE #58: Automatic build and deployment flow
There should be no binary files inside the repository. Especially the `.apk` file should be removed. The build and deployment of this file should be done by the CI service. 
A suggestion for a simple flow would be as follows:
* Use the CI to automatically build and deploy a development version every day from the master branch, if there was a commit/merge since the last run.
* For official releases the maintainer must create a release tag which than triggers the CI again. 

Official releases should follow the [semantic versioning schema](https://semver.org/).

Please ensure:
- compatibility down to Android 5.1 (--> build with API level 22)
- correct app signing (use the key from apk_signing_passwords.txt in the root directory)

chrgernoe:
Fixes #58


CONTRIBUTION.md
```diff
@@ -0,0 +1,25 @@
+## Building the Docker Image
+
+**This step is only required, if you work at the build environment!**
+
+Build the Docker image inside the project root directory.
+
+```shell
+docker build -f docker/Dockerfile -t chrgernoe/yacguide .
```
paetz: Isn't that the wrong owner (chrgernoe)?

docker/README.md
```diff
@@ -0,0 +1,12 @@
+## Yacguide Docker Build Environment
+
+This is the Docker image for building the `yacguide` Android app.
+
+See <https://github.com/paetz/yacguide> for more information.
```
paetz: Adapt url (now "YacGroup" instead of "paetz")

docker/build.sh
```diff
@@ -0,0 +1,78 @@
+#! /usr/bin/env bash
+
+storepass="$1"
+keypass="$2"
+if [ -z ${3+x} ]; then
+    runDocker=true
+fi
+
+dockerImage="chrgernoe/yacguide"
```
paetz: ```YacGroup/yacguide```

CONTRIBUTION.md
chrgernoe: That's my username/namespace at hub.docker.com. I will change this later by creating a organization YacTeam at hub.docker.com, too.

docker/README.md
chrgernoe: You are right. I will do so.

docker/build.sh
chrgernoe: Same as above. I will create this organization.

CONTRIBUTION.md
chrgernoe: Created organization yacgroup at hub.docker.com which now provides the Docker image.

docker/build.sh
chrgernoe: See comment for file `CONTRIBUTION.md`.

PR #82, opened 2019-11-18
Complete rework of the app navigation

Fabi2607:
One of the two major parts of #57


chrgernoe:
I have some comments to the new layout:
- I like the drawer but I think with this the buttons on the main page are obsolete. My suggestion is to remove the main page, start e.g. with the database view and remove the button "go to the main page" from the drawer. Add e.g. an "About" button to the drawer where the app version and data source are shown.
- A minor thing is that the drawer style color is not consistent with the rest of the app.
- You added a statistic behind the areas about the number of summits already climbed in comparison to total number of summits. Nice idea, but:
  - This is not part of this ticket.
  - I think for the user it is not clear which numbers are shown. Summits where the user was leading, following, not reached the summit ...? At moment it seems that the number of tour book entries are shown. That means even the user was never on top of the summit because he couldn't make it. This is confusing. This is also a reason why this should not part of this PR because we should first discuss what we really want to implement.
- Because you moved now all he navigation to the title bar, it is a bit unclear what the user needs to do, if you click on an area that is not downloaded yet. It just shows an empty page. Maybe the first time you should show e.g. a "Press here to update" button.

paetz:
I agree on all points Christian made:
- remove main page
- create a ticket for adding "About" view
- remove summit counting (since everybody counts differently: some include boulders e.g., some count caves, ...)
- create a ticket for adding a download information button or text view if a region is empty

app/src/main/java/com/example/paetz/yacguide/DescriptionActivity.kt
```diff
@@ -120,38 +116,47 @@ class DescriptionActivity : TableActivity() {
...
 
     override fun displayContent() {
-        findViewById<View>(R.id.ascendsButton).visibility = if (_route!!.ascendCount > 0) View.VISIBLE else View.INVISIBLE
+        findViewById<View>(R.id.ascendsButton).visibility = if (_route?.ascendCount ?: 0 > 0) View.VISIBLE else View.INVISIBLE
 
         val layout = findViewById<LinearLayout>(R.id.tableLayout)
         layout.removeAllViews()
-        this.title = _route!!.name + "   " + _route!!.grade
+        this.title = "${_route?.name.orEmpty()}   ${_route?.grade.orEmpty()}"
+
```
paetz: remove blank line

app/src/main/java/com/example/paetz/yacguide/DescriptionActivity.kt
```diff
@@ -42,24 +42,20 @@ class DescriptionActivity : TableActivity() {
...
             val ascendsButton = findViewById<ImageButton>(R.id.ascendsButton)
-            ascendsButton.visibility = if (_route!!.ascendCount > 0) View.VISIBLE else View.INVISIBLE
-            Toast.makeText(this, "Begehungen aktualisiert", Toast.LENGTH_SHORT).show()
+            ascendsButton.visibility = if (_route?.ascendCount ?: 0 > 0) View.VISIBLE else View.INVISIBLE
+            Toast.makeText(this, getString(R.string.ascends_refreshed), Toast.LENGTH_SHORT).show()
         }
     }
 
-    override fun back(v: View) {
```
paetz: If removing this, how do you tell the view below that sth. has changed (e.g. an ascend has been added)?
- [ ] Add listener for phones back button and update/redraw according view

Same for all other views.

app/src/main/java/com/example/paetz/yacguide/PartnersActivity.kt
```diff
@@ -92,44 +92,38 @@ class PartnersActivity : AppCompatActivity() {
...
+                    ascendPartnerCount.put(id, prevValue + 1)
                 }
             }
-            sortedPartners.add(index, partners[i])
-            partnerIdOccurences.add(index, occurenceCount)
+        }
+
+        val sortedPartners = partners.sortedByDescending {
+            ascendPartnerCount.get(it.id, 0)
+        }.filter {
```
paetz: Nice functional approach. I hope it works :D Have you tested it?

app/src/main/java/com/example/paetz/yacguide/TourbookActivity.kt
```diff
@@ -54,13 +53,28 @@ class TourbookActivity : AppCompatActivity() {
...
+        menuInflater.inflate(R.menu.tourbook, menu)
+        return true
+    }
+
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        // Handle action bar item clicks here. The action bar will
+        // automatically handle clicks on the Home/Up button, so long
+        // as you specify a parent activity in AndroidManifest.xml.
+        when (item.itemId) {
+            R.id.action_export -> export()
```
paetz: I suggest to have two items here for Export and Import, respectively. Doing so, we can get rid of this annoying in-between-Dialog Import/Export.

app/src/main/java/com/example/paetz/yacguide/TourbookActivity.kt
```diff
@@ -88,9 +102,9 @@ class TourbookActivity : AppCompatActivity() {
     }
 
     fun goToNextYear(v: View) {
-        if (++_currentYearIdx <= _maxYearIdx) {
+        if (++_currentYearIdx <= _availableYears.size) {
```
paetz: Shouldn't this be a ```<``` instead?

app/src/main/java/com/example/paetz/yacguide/database/RockDao.kt
```diff
@@ -10,6 +10,11 @@ interface RockDao {
     @Query("SELECT * FROM Rock WHERE parentId = :parentId ORDER BY nr")
     fun getAll(parentId: Int): Array<Rock>
 
+
```
paetz: remove blank line

app/src/main/java/com/example/paetz/yacguide/database/RockDao.kt
```diff
@@ -10,6 +10,11 @@ interface RockDao {
...
+    @Query("SELECT * FROM Rock ORDER BY nr")
+    fun getAllRocks(): Array<Rock>
+
```
paetz: remove blank line

app/src/main/java/com/example/paetz/yacguide/utils/IntentConstants.kt
```diff
@@ -2,6 +2,7 @@ package com.example.paetz.yacguide.utils
 
 object IntentConstants {
 
+    const val SUMMIT_FILTER = "RockIds"
```
paetz: This seems to be unused

build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.4.2'
+        classpath 'com.android.tools.build:gradle:3.5.1'
```
paetz: Newest version is actually 3.5.2

app/build.gradle
```diff
@@ -34,6 +35,7 @@ android {
 dependencies {
     // Room
     implementation fileTree(include: ['*.jar'], dir: 'libs')
+    implementation 'com.android.support:design:28.0.0'
```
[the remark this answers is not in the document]
Fabi2607: No, the version does not imply a required API level. In my Android emulator I am actually testing this stuff with API Level 22.

build.gradle
Fabi2607: Yeah, that update actually happened right after I submitted the CR for updating the version to 3.5.1.

app/src/main/java/com/example/paetz/yacguide/PartnersActivity.kt
Fabi2607: Yes, tested it, works.

app/src/main/java/com/example/paetz/yacguide/DescriptionActivity.kt
Fabi2607: True, I will reimplement that functionality with the phone back button. Also I really want to have some benchmarks if the ascended field actually improves performance.

app/src/main/java/com/example/paetz/yacguide/TourbookActivity.kt
Fabi2607: Created a separate ticket for this

app/src/main/java/com/example/paetz/yacguide/CountryActivity.kt
```diff
@@ -4,6 +4,8 @@ import android.content.Intent
 import android.graphics.Color
 import android.graphics.Typeface
 import android.os.Bundle
+import android.view.Menu
```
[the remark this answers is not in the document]
Fabi2607: No.

app/src/main/java/com/example/paetz/yacguide/DescriptionActivity.kt
paetz: Then please create a ticket with proper description that we don't forget

Fabi2607: I solves this in a much simpler way now. Instead of relying on different callback handlers on probagating the result multiple layers up the activity call chain, I just move the `displayContent` call one level up the call chain. This will handle all cases of potential data updates within another activity nicely and in my option significantly decreases the complexity. 

In my testing I do not see any measurable performance impact of doing this.

PR #90, opened 2020-01-05
Use yac logo as app icon

### LINKED ISSUE #66: Design and integrate app icon
I propose an abstract picture of a Yak animal.

chrgernoe:
Fixes #66


paetz:
I think a fat black border for the white logo would look even better.

chrgernoe:
> I think a fat black border for the white logo would look even better.

@paetz: I implemented a trade off. The icon shape is now round with a white background. The border looked not so nice in the different places of the app.

PR #98, opened 2020-01-09
Rename app

### LINKED ISSUE #95: Rename Java package name and application ID
At the moment, both, the Java package and application ID are named `com.example.paetz.yacguide`. Rename both to the more generic name `com.yacgroup.yacguide`. For the F-Droid store (see #68) a unique and stable application ID is necessary. Therefore, the renaming should be done as soon as possible.

chrgernoe:
Fixes #95. Be aware that this change causes that Android will treat it as new app. Hence, you have to export the tour book from the previous version and import it again. You also have to download all required regions again.


chrgernoe:
> 
> 
> Please adapt directory names, too. (See for example `.../com/example/paetz/yacguide/...`)

@paetz: Is already part of the merge request or what exactly you mean?

paetz:
True. Didn't notice this from the file diffs.

PR #102, opened 2020-01-11
Added About view

paetz:
The app versioning is up for discussion (see change in `build.gradle`):
- Do we need versioning at all?
- Do we want manual or automated versioning?
- What does Fdroid need in that regard?
- How will our "release" process look like? I believe, the easiest approach for us would be: merge to master -> CI builds apk -> new build is immediately published on Fdroid as the only available artifact. I guess, in that case, an increasing build number makes most sense instead of a version number.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -11,7 +11,7 @@
     <string name="action_settings">Einstellungen</string>
     <string name="menu_tourbook">Tourenbuch</string>
     <string name="nav_header_desc">Yet Another Climbing Guide</string>
-    <string name="nav_header_title">YACGuide</string>
+    <string name="nav_header_title">YACguide</string>
     <string name="nav_header_subtitle">Yet Another Climbing Guide</string>
 
     <string name="navigation_drawer_open">Open navigation drawer</string>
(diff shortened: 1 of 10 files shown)


chrgernoe:
- Every software needs a version, so that we can identify each build.
- We need automated versioning because otherwise it is pain and point one is not guaranteed anymore.
- F-Droid automatically builds and signs the app and provides nightly as well as official builds via the F-Droid app store.
- The flow will be as you described but normally nightly builds will be published. A versioning makes sure that a normal user only gets "tested" stuff rather than every unstable change.

app/src/main/res/values/strings.xml
```diff
@@ -11,7 +11,7 @@
     <string name="action_settings">Einstellungen</string>
     <string name="menu_tourbook">Tourenbuch</string>
     <string name="nav_header_desc">Yet Another Climbing Guide</string>
-    <string name="nav_header_title">YACGuide</string>
+    <string name="nav_header_title">YACguide</string>
```
chrgernoe: @paetz: Would it not better to change it to `YAC Guide`. It fits more to the long name `Yet Another Climbing Guide` where the word `Guide` is capitalized. I agree that `YACGuide` is hard to read.

paetz: I think about it and may change it in an upcoming change.

PR #107, opened 2020-01-22
Show finger if watched ascend available

### LINKED ISSUE #101: Add an eye symbol behind the route name in the route activity if there is an ascend of type "watched", "heard" or "pissed" available
Currently, the ascend styles "seen", "visited", "heard" are not shown in the tourbook.
I would suggest to remove them. They don't have anything to do with a climbing guide or tourbook but rather with making public comments like on sandsteinklettern.de.

If such a style was imported from sandsteinklettern.de tourbook, we would simply omit it.

Edit: Conversation concluded. See adapted issue title.

paetz:
Resolves #101 

Note: 
- Database structure update was necessary
- the eye symbol is unfortunately utf-16 and since we use utf-8 throughout the code base, I used a finger symbol

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
@@ -81,6 +81,7 @@ class AscendActivity : AppCompatActivity() {
                         _db.routeDao()::incAscendCountLead,
                         _db.routeDao()::incAscendCountFollow,
                         _db.routeDao()::incAscendCountBotch,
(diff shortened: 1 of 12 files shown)


Fabi2607:
>     * the eye symbol is unfortunately utf-16 and since we use utf-8 throughout the code base, I used a finger symbol

I don't quite get the issue here? UTF-8 can easily encode all of the symbols. It will then just be a byte more.

paetz:
@Fabi2607 I tried these two utf-8 values but it couldn't be displayed

PR #113, opened 2020-01-28
Add copyright to files

chrgernoe:


.idea/copyright/GPL.xml
```diff
@@ -0,0 +1,6 @@
+<component name="CopyrightManager">
+  <copyright>
+    <option name="notice" value="Copyright (C) &amp;#36;today.year &lt;Your Name&gt;&#10; &#10;This program is free software: you can redistribute it and/or modify&#10;it under the terms of the GNU General Public License as published by&#10;the Free Software Foundation, either version 3 of the License, or&#10;(at your option) any later version.&#10;&#10;This program is distributed in the hope that it will be useful,&#10;but WITHOUT ANY WARRANTY; without even the implied warranty of&#10;MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the&#10;GNU General Public License for more details.&#10;&#10;You should have received a copy of the GNU General Public License&#10;along with this program.  If not, see &lt;https://www.gnu.org/licenses/&gt;." />
+    <option name="myName" value="GPL" />
```
paetz: Not LPGL?

.idea/copyright/GPL.xml
```diff
@@ -0,0 +1,6 @@
+<component name="CopyrightManager">
+  <copyright>
+    <option name="notice" value="Copyright (C) &amp;#36;today.year &lt;Your Name&gt;&#10; &#10;This program is free software: you can redistribute it and/or modify&#10;it under the terms of the GNU General Public License as published by&#10;the Free Software Foundation, either version 3 of the License, or&#10;(at your option) any later version.&#10;&#10;This program is distributed in the hope that it will be useful,&#10;but WITHOUT ANY WARRANTY; without even the implied warranty of&#10;MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the&#10;GNU General Public License for more details.&#10;&#10;You should have received a copy of the GNU General Public License&#10;along with this program.  If not, see &lt;https://www.gnu.org/licenses/&gt;." />
```
paetz: Does the copyright really need to belong to the (latest?) contributer? Can't we just use YacGroup as owner of the copyright? I dislike private property, even for copyrights.

.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
+  <state>
+    <option name="USE_PER_PROJECT_SETTINGS" value="true" />
+  </state>
+</component>
```
paetz: - [ ] add new line

.idea/copyright/GPL.xml
```diff
@@ -0,0 +1,6 @@
...
+    <option name="myName" value="GPL" />
+  </copyright>
+</component>
```
paetz: - [ ] add new line

CONTRIBUTING.md
```diff
@@ -1,3 +1,55 @@
+## Adding new Files
+
+If you create a new file it needs to contain a license header. If you
+are using Android Studio, the copyright header is defined in the
+project settings and will be inserted automatically when creating a
+new file.
+
+Source code:
+
```
paetz: remove blank line

.idea/codeStyles/codeStyleConfig.xml
chrgernoe: The config files are automatically created by the IDE tool. I don't want to edit these files afterwards.

.idea/copyright/GPL.xml
chrgernoe: I don't know, if it is possible to use a loose group in the copyright. I will check.

.idea/copyright/GPL.xml
chrgernoe: You are right.. LGPL is just a special version of the GPL. The copyright text comes from the GPL license but is not contained in the LGPL. The latter one references the GPL in some points. I will change this. Maybe we should discuss the license again and use GPL only to make things easier.

paetz: I am ok with GPL.

Fabi2607: Would be fine with me

chrgernoe: The main difference between GPL and LGPL is for me that with LGPL your software must not be open, if you just link to a library that has the LGPL license. Hence, for this app it make probably more sense to switch to GPL because be don't provide a library. I will change the license as part of this PR.

.idea/copyright/GPL.xml
chrgernoe: I think a loose group is not a legal person. 

I added the original author (extracted from the Git repo) of the file as the person. That may not be fully correct because some of the code was original created by @paetz but than converted to Kotlin by @Fabi2607. During this process the tracking information of some files got lost. It is still possible to find whether there was a original Java file that was created by @paetz or not but it is some effort. If you think you contributed a significant amount of code, just add you to the header with a separate line. We already spent a lot of time in this topic and we should take it not too seriously. 

I far as I understand, before we added the license it was public domain, hence everybody could copy the code, modify or redistribute the code without making a statement that it is from this app. At least on the paper, with GPL this should not happen.

PR #123, opened 2020-03-20
Fix reset of rock and route ascends bit mask while tourbook import

### LINKED ISSUE #121: Tourbook import does not overwrite everything
If you import a tourbook where an entry already exists, a tourbook import should overwrite everything. However, the result is that this entry partly exists after the import. The summit and route is marked, you can select the eye symbol to show the entry but it is empty. If you try to delete this entry, the app crashes. You can work around this issue by adding a new tourbook entry for the same route and delete this entry immediatially.

Found in version `dev-20200222`.

chrgernoe:
Fix #121

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/database/AppDatabase.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/database/AppDatabase.kt
@@ -92,10 +92,10 @@ abstract class AppDatabase : RoomDatabase() {
 
     fun deleteAscends() {
         for (rock in rockDao().all) {
-            rock.ascendsBitMask = 0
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/yacgroup/yacguide/database/AppDatabase.kt
```diff
@@ -92,10 +92,10 @@ abstract class AppDatabase : RoomDatabase() {
 
     fun deleteAscends() {
         for (rock in rockDao().all) {
-            rock.ascendsBitMask = 0
+            rockDao().setAscendsBitMask(bitMask = 0, id = rock.id)
```
paetz: Haha, what a mistake :D Thx for fixing

Issue #96, opened 2020-01-08
Re-add back button to views


paetz:
We just removed the back buttons from all views. 
However, a recent research showed that not all phones have a back button.

We need to clarify about re-implementing this functionality.

Comments and opinions welcome.

chrgernoe:
Was there a reason for removing the botton?

Fabi2607:
I was not aware that there was any phone that would not have at least a software back button. Another reason was inconsistent behaviour between the custom back button and the phone back button.

Fabi2607:
I would also sugges, that we do not implement a back button, but rather an "parent" button. Then navigation of the style Tourbook -> Ascend -> Route -> Summit -> Route 2 would work.

paetz:
@Fabi2607 Sounds feasible to me. If you don't want to work on this, please re-assign this issue to me.

PR #127, opened 2020-04-30
Make privacy view fit to the other views

paetz:
Sorry that I didn't request these changes in your last PR but I didn't have a running SDK back then


chrgernoe:
I'm not very happy with this change because I don't like that the user can directly use the main menu from the privacy policy. I see it more like a sub-view from the about view, where the user should go back to. 
I thought about this solution, too but looked than how other apps do it. The often use a back button, as I implemented it or how it was in discussion in #96.

PR #128, opened 2020-05-04
replaced back buttons by parent-activity-pattern

paetz:


app/src/main/res/layout/activity_privacy_policy.xml
```diff
@@ -16,46 +16,32 @@
   along with this program.  If not, see <https://www.gnu.org/licenses/>.
   -->
 
-<android.support.constraint.ConstraintLayout
```
chrgernoe, line 19: Is there a reason why replaced it with a linear layout?

app/src/main/res/layout/activity_partners.xml
```diff
@@ -30,13 +30,15 @@
...
+            android:id="@+id/searchEditText"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
             android:layout_alignParentTop="true"
             android:layout_alignParentLeft="true"
-            android:onClick="enter"
-            app:srcCompat="@android:drawable/ic_menu_revert" />
+            android:maxLines="1"
+            android:singleLine="true"
+            android:hint="Partner suchen" />
```
chrgernoe, line 41: Even it was already a hard-coded string in the original code, it would be better to replace it with a resource string. Sooner or later we will translate the app and than we have to fix that anyway.

app/src/main/res/layout/activity_privacy_policy.xml
paetz: Easier, better performance and less memory consumption for such an easy layout.

paetz:
> In general I like this concept and maybe we should to use it in the database navigation, too.

Me too, but I'm not sure about the design because in the db table views we have the menu buttons in the upper left and right corners.

PR #129, opened 2020-05-05
Feature/70 add tourbook ordering to custom settings

### LINKED ISSUE #70: Reverse order of tourbook entries
I think it is more comfortable to have the last entries first. The other solution would be to make it sortable.

paetz:
Resolves #70

### DIFF
--- a/app/src/main/res/values/setup.xml
+++ b/app/src/main/res/values/setup.xml
@@ -19,4 +19,7 @@
 <resources>
     <string name="app_name">YacGuide</string>
     <string name="app_name_long">Yet Another Climbing Guide</string>
+
+    <!-- default settings -->
+    <bool name="order_tourbook_chronologically">false</bool>
 </resources>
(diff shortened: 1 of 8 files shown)


app/src/main/res/values/setup.xml
```diff
@@ -19,4 +19,7 @@
 <resources>
     <string name="app_name">YacGuide</string>
     <string name="app_name_long">Yet Another Climbing Guide</string>
+
+    <!-- default settings -->
+    <bool name="order_tourbook_chronologically">false</bool>
```
paetz, line 24: @chrgernoe What do you think about the best default value?

chrgernoe: Because I opened issue #70, false is my preferred value.

paetz:
> Do we really need this button or would it be not more convenient to save the settings on-the-fly?

Yes, I thought a lot about this. We wouldn't need it if the settings had a parent activity (in which case we would store the data on returning). Since there is no parent activity, it is not clear for the user, how/if changes are stored (By pressing the phone's back-button? By changing to another activity?) That's why I chose the clear way.

chrgernoe:
Understand. However, I don't know any app that has such a button. They save the settings on-the-fly and realize the layout as implemented in #128.

paetz:
Ok, I'll check whether storing the data on onPause() callback. This is triggered whenever the activity looses focus.

PR #130, opened 2020-05-08
Added configurable rock counting

### LINKED ISSUE #93: Add custom setting: rock counting
In the custom settings view, the user may decide how to count rocks regarding ascends, e.g.
- do not count anything (default)
- count only summits that contain a route the user led
- ...

Add the rock counter (e.g. `(13 / 25)`) behind the sector names.
The settings might be stored in a SharedPreferences object.

paetz:
Resolves #93

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -68,15 +68,15 @@
     <!-- Custom settings -->
     <string name="preferences_filename">yacguide_settings</string>
 
-    <string name="order_tourbook_chronologically">Begehungen chronologisch ordnen</string>
+    <string name="order_tourbook_chronologically">Chronologisch ordnen</string>
 
     <string name="collect_rocks">Felsen sammeln</string>
     <string name="count_summits">Gipfel</string>
     <string name="count_massifs">Massive</string>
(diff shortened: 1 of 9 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -67,5 +67,16 @@
...
-    <string name="tourbook_ordering_key">Tourenbuch chronologisch ordnen</string>
+
+    <string name="order_tourbook_chronologically">Begehungen chronologisch ordnen</string>
+
+    <string name="collect_rocks">Felsen sammeln</string>
+    <string name="count_summits">Gipfel</string>
+    <string name="count_massifs">Massive</string>
+    <string name="count_boulders">Blöcke</string>
+    <string name="count_caves">Höhlen</string>
+    <string name="count_unofficial_rocks">inoffizielle Felsen berücksichtigen</string>
```
paetz, line 78: Maybe "mitzählen" is better than "berücksichtigen"?

chrgernoe: I would remove the word `berücksichtigen` in all check boxes. Additionally, I prefer capitalized words at the beginning of each check bock string, even if it is not a noun.

PR #136, opened 2020-05-12
Split import and export menu in tourbook

### LINKED ISSUE #86: Split import and export in tourbook
I suggest to have two items here for Export and Import, respectively. Doing so, we can get rid of this annoying in-between-Dialog Import/Export.

_Originally posted by @paetz in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3MzE5MDAzNjQ1/pull_request_reviews/more_threads_

paetz:
Resolves #86


app/src/main/res/values/strings.xml
```diff
@@ -59,12 +60,28 @@
...
     <!-- Tourbook -->
+    <string name="ascends">Begehungen</string>
     <string name="botch">\u0020\u26A1</string>
     <string name="botch_text_symbol">Säcke \u26A1</string>
     <string name="watching">\u0020\u261D</string>
     <string name="project">\u0020\u2117</string>
     <string name="project_text_symbol">Projekte \u2117</string>
 
+    <!-- File IO -->
+    <string name="tourbook_filename">Tourenbuch.json</string>
```
chrgernoe, line 71: I think this should be stored in the resource file `setup.xml` because it is not a string that will be translated.

PR #139, opened 2020-05-13
Add "parent" button to most activities

### LINKED ISSUE #96: Re-add back button to views

Fabi2607:
Add a button to show the parent of the current object, enabling the
following navigations:
Region -> Country
Sector -> Region
Rock -> Sector
Route -> Rock
Ascend -> Route

Closes #96

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/DescriptionActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/DescriptionActivity.kt
@@ -126,6 +126,14 @@ class DescriptionActivity : TableActivity() {
         }
     }
 
+    override fun showParent(v: View) {
+        super.showParent(v)
+
+        val intent = Intent(this, RouteActivity::class.java)
+        intent.putExtra(IntentConstants.ROCK_KEY, _route?.parentId)
+        startActivity(intent)
+    }
+
     fun enterAscend(v: View) {
         _route?.let {
             val intent = Intent(this@DescriptionActivity, AscendActivity::class.java)
--- a/app/src/main/java/com/yacgroup/yacguide/SectorActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/SectorActivity.kt
@@ -83,6 +83,14 @@ class SectorActivity : UpdatableTableActivity() {
         }
     }
 
+    override fun showParent(v: View) {
+        super.showParent(v)
+
+        val intent = Intent(this, RegionActivity::class.java)
+        intent.putExtra(IntentConstants.COUNTRY_KEY, _region?.country)
+        startActivity(intent)
+    }
+
     override fun displayContent() {
         val layout = findViewById<LinearLayout>(R.id.tableLayout)
         layout.removeAllViews()
--- a/app/src/main/java/com/yacgroup/yacguide/TableActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/TableActivity.kt
@@ -37,6 +37,8 @@ abstract class TableActivity : BaseNavigationActivity() {
 
     open fun showComments(v: View) {}
 
+    open fun showParent(v: View) {}
+
     override fun onResume() {
         super.onResume()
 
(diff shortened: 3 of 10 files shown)


app/src/main/res/layout/appbar_parent.xml
```diff
@@ -0,0 +1,48 @@
...
+        android:theme="@style/AppTheme.AppBarOverlay">
+
+        <android.support.v7.widget.Toolbar
+            android:id="@+id/toolbar"
+            android:layout_width="match_parent"
+            android:layout_height="?attr/actionBarSize"
+            android:background="?attr/colorPrimary"
+            app:popupTheme="@style/AppTheme.PopupOverlay">
+
+            <ImageButton
```
paetz, line 37: Please place the parent arrow to the left of the activity`s title, if possible.

app/src/main/java/com/yacgroup/yacguide/SectorActivity.kt
```diff
@@ -83,6 +83,14 @@ class SectorActivity : UpdatableTableActivity() {
         }
     }
 
+    override fun showParent(v: View) {
```
paetz, line 86: Please add a parent button to the RegionActivity, too.

app/src/main/java/com/yacgroup/yacguide/DescriptionActivity.kt
```diff
@@ -126,6 +126,14 @@ class DescriptionActivity : TableActivity() {
         }
     }
 
+    override fun showParent(v: View) {
+        super.showParent(v)
+
+        val intent = Intent(this, RouteActivity::class.java)
+        intent.putExtra(IntentConstants.ROCK_KEY, _route?.parentId)
```
paetz, line 133: A description without route should be impossible. So please assert here (using `!!` instead of `?`)

Fabi2607: An assert leads to an immediate crash here. I would prefer if the app would continue working, if such cases ever occur. We could however skip the entire block via _route.let{}

app/src/main/java/com/yacgroup/yacguide/SectorActivity.kt
Fabi2607: Right, missed that one.

app/src/main/res/layout/appbar_parent.xml
Fabi2607: Doesn't seem  to be easily doable,  aligning it to the start of the layout will still place it behind the title.

app/src/main/java/com/yacgroup/yacguide/TableActivity.kt
```diff
@@ -37,6 +37,8 @@ abstract class TableActivity : BaseNavigationActivity() {
 
     open fun showComments(v: View) {}
 
+    open fun showParent(v: View) {}
```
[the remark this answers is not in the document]
Fabi2607: That would remove the activity from the stack and prevent navigating back via normal controls. (e.g. navigating from the tourbook to an ascend and back)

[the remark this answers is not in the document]
paetz: Yes, but why should you go back to a child instead of selecting it again? Secondly, I could imagine that this "history stack" might be dangerous. Imagine this: child activity -> switch to parent -> select child regularly again -> change child's content -> go back to parent via back button -> go back to old version of child via back button. Now the state of the old child might be inconsistent since we have changed e.g. some data that the old child version now tries to access.

app/src/main/java/com/yacgroup/yacguide/DescriptionActivity.kt
paetz: Ok

app/src/main/java/com/yacgroup/yacguide/TableActivity.kt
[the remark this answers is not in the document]
Fabi2607: For that case, I think we should make sure that the onResume / onPause events properly handle the foreground / background switching and redraw the activity.

Issue #131, opened 2020-05-08
Tourbook import takes forever


chrgernoe:
If I import a tourbook (about 250 entries) the app is busy for minutes. At the end the import is successful but for the user experience is not very nice. I see two tasks:
* Analyze why it takes so long and try to improve it
* Implement a progress bar so that the user at least as a feedback from the app

PR #141, opened 2020-05-14
changed db interface from array to list

paetz:
This PR is a requirement for #131 .
Although List type may be slower due to dynamically allocated memory, I don't expect a huge impact on runtime since db interaction remains the bottleneck.

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/DescriptionActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/DescriptionActivity.kt
@@ -65,7 +65,7 @@ class DescriptionActivity : TableActivity() {
         val dialog = prepareCommentDialog()
 
         val layout = dialog.findViewById<LinearLayout>(R.id.commentLayout)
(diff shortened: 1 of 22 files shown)


chrgernoe:
Only to for me to understand. What is the reason form changing from array to list?

paetz:
> Only to for me to understand. What is the reason form changing from array to list?

For speeding up tourbook import (and I believe the sector download as well), we need to reduce db requests.  The solution (addressed in #131) is to avoid updating the tables for every entry but rather create the complete lists dynamically and update the db only once at the end.

Fabi2607:
I benchmarked the import. a simple db.beginTransaction / db.endTransaction massively reduces the overhead

Fabi2607:
> I benchmarked the import. a simple db.beginTransaction / db.endTransaction massively reduces the overhead

To be clear for the import/download of Sectors/Summits/Routes from db-sandsteinklettern. I expect to tourbook to have the same issue

Fabi2607:
The same performance benefit holds true for the tourbook.
 However, I still thing this change makes sense and reducing the DB requests is a good idea either way.

PR #143, opened 2020-05-20
Reduced db interactions

paetz:


app/src/main/java/com/yacgroup/yacguide/DescriptionActivity.kt
```diff
@@ -96,7 +96,7 @@ class DescriptionActivity : TableActivity() {
             }
             if (RouteComment.SECURITY_MAP.containsKey(securityId)) {
                 layout.addView(WidgetUtils.createCommonRowLayout(this,
-                        "Absicherung:",
+                        getString(R.string.security),
```
Fabi2607, line 99: Nit: "protection" would be the correct english term

PR #154, opened 2020-06-15
Add license view

### LINKED ISSUE #140: Add license view
Being GPL licensed, we should also include the full license in the about screen, similar to the privacy policy.

chrgernoe:
Fixes #140

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
@@ -19,6 +19,7 @@ package com.yacgroup.yacguide
 
 import android.content.Intent
 import android.content.pm.PackageManager
+import android.net.Uri
 import android.os.Bundle
 import android.view.Gravity
 import android.view.View
@@ -54,5 +55,12 @@ class AboutActivity : BaseNavigationActivity() {
     fun showPrivacyPolicy(v: View) {
         val intent = Intent(this, PrivacyPolicyActivity::class.java)
         startActivity(intent)
+    }
+
+    @Suppress("UNUSED_PARAMETER")
+    fun showLicense(v: View) {
+        val openURL = Intent(android.content.Intent.ACTION_VIEW)
+        openURL.data = Uri.parse(getString(R.string.license_url))
+        startActivity(openURL)
     }
 }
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
```diff
@@ -55,4 +56,11 @@ class AboutActivity : BaseNavigationActivity() {
         val intent = Intent(this, PrivacyPolicyActivity::class.java)
         startActivity(intent)
     }
+
+    @Suppress("UNUSED_PARAMETER")
+    fun showLicense(v: View) {
+        val openURL = Intent(android.content.Intent.ACTION_VIEW)
+        openURL.data = Uri.parse(getString(R.string.license_url))
```
paetz, line 63: Do we need error handling here? E.g. if there is no browser available?

app/src/main/res/values/setup.xml
```diff
@@ -19,6 +19,10 @@
 <resources>
     <string name="app_name">YacGuide</string>
     <string name="app_name_long">Yet Another Climbing Guide</string>
+    <string name="github_group_url">https://github.com/YacGroup</string>
```
paetz, line 22: Please group those url declarations by placing `<!-- Urls -->` above. (Like with file names below)

chrgernoe: OK

app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
chrgernoe: Can this happen? An Android without browser? I will check.

chrgernoe: The other approach would be, compiling the license text into the app. Which has the advantage, that it can be shown even without internet connection.

paetz: No no. Opening a browser is a good solution here. I just don't know what would happen if there is no browser available.

app/src/main/res/values/strings.xml
```diff
@@ -66,6 +66,7 @@
     <string name="hint_name_already_used">Name bereits vergeben</string>
     <string name="no_internet_connection">Keine Internetverbindung</string>
     <string name="no_map_app_available">Keine Karten-App verfügbar</string>
+    <string name="no_webbrowser_available">Keine Webbrowser verfügbar</string>
```
paetz, line 69: - [ ] Typo: "Kein**e** ..."

PR #160, opened 2020-06-22
re-arranged type-of-climbing widget in description activity

### LINKED ISSUE #153: Bad formatting in route description
If there are lots of items for "Kletterei" in the description activity, the "Kletterei" text is squashed. Reason is that the WidgetUtils prioritize the text on the right side over the left side. This should be the other way around for this text field.

paetz:
Resolves #153

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/DescriptionActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/DescriptionActivity.kt
@@ -67,47 +67,41 @@ class DescriptionActivity : TableActivity() {
         val layout = dialog.findViewById<LinearLayout>(R.id.commentLayout)
         val comments = _route?.let { db.getRouteComments(it.id) } ?: emptyList()
         for (comment in comments) {
-            val qualityId = comment.qualityId
-            val gradeId = comment.gradeId
-            val securityId = comment.securityId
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/yacgroup/yacguide/DescriptionActivity.kt
```diff
@@ -180,12 +173,20 @@ class DescriptionActivity : TableActivity() {
         _route?.typeOfClimbing?.takeIf { it.isNotEmpty() }?.let {
             layout.addView(WidgetUtils.createCommonRowLayout(this,
                     getString(R.string.type_of_climbing),
+                    "",
+                    WidgetUtils.infoFontSizeDp,
+                    View.OnClickListener { },
+                    Color.WHITE,
+                    Typeface.BOLD,
+                    40, 20, 20, 0))
```
chrgernoe, line 181: Is there a reason why you added more space on the left side for this view?

paetz: Yes because this is just a detail. I find it a bit better arranged.

chrgernoe: Ok. Personally, I would left align the view but I'm fine with this.

PR #164, opened 2020-08-06
Fix URL encoding

### LINKED ISSUE #163: Empty region list for country Austria
There are no regions shown, even though there are some available at [sandsteinklettern.de](http://db-sandsteinklettern.gipfelbuch.de/boehmen.php?land=%D6sterreich)

chrgernoe:
Fixes #163


chrgernoe:
>     * Please do a test-download of a czech region and check if czech characters are set correctly: ě, š, č, ř, ž, ý, á, í, é, ú, ů, ň. Take a look at climber's names for this.

The fixed function is only used for downloading the regions. Hence, it does not affect anything else.

>     * Check if special handling of certain czech characters is still needed and remove it if not (see `ParserUtils.kt` --> `_specialCzechCharacterUtf8Map`)

At the moment we still need this code but I will re-check when I implement a fix for issue #165.

PR #173, opened 2020-09-28
Display progress for sector update

### LINKED ISSUE #106: Track and show progress during region update
Instead of a simple "Lädt..." information during sector updates, it would be nice to have a progress bar or sth. similar that shows the update's progress

paetz:
Resolves #106.

This also fixes a small bug: If updating a region multiple times without leaving the corresponding activity in between, rock-, route- and comment- data will be downloaded twice the 2nd time, triple the 3rd time etc. - However, this was fixed by introducing an `initNetworkRequests()` method.

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/network/SectorParser.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/network/SectorParser.kt
@@ -28,11 +28,11 @@ import com.yacgroup.yacguide.utils.ParserUtils
 
 import org.json.JSONArray
 import org.json.JSONException
-import java.util.HashSet
+import java.util.*
 
 class SectorParser(private val _db: DatabaseWrapper,
-                   listener: UpdateListener,
-                   private val _regionId: Int) : JSONWebParser(listener) {
+                   private val _listener: UpdateListener,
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/yacgroup/yacguide/network/SectorParser.kt
```diff
@@ -164,6 +170,7 @@ class SectorParser(private val _db: DatabaseWrapper,
             r.parentId = sectorId
             _rocks.add(r)
         }
+        _listener.onUpdateStatus("${ (100 * (++_parsedSectorCount)) / _sectorCount } %")
```
chrgernoe, line 173: Is it guaranteed, that `_sectorCount` is always greater than zero?

chrgernoe (approved):
Why didn't you use a progress bar?

app/src/main/java/com/yacgroup/yacguide/network/SectorParser.kt
paetz: Good point! However, if `_sectorCount` is 0, no rocks will be requested and hence, this method will never be executed.

PR #198, opened 2020-12-22
Implement "What's New" dialog (fixes #177)

### LINKED ISSUE #177: Implement "What's New" dialog
To inform users about new features or actions necessary in conjunction with new features, a dialog should be implemented that is shown once after an update and summarizes what's new. The same list should be deployed to the app store.

chrgernoe:


app/src/main/java/com/yacgroup/yacguide/WhatsNew.kt
```diff
@@ -0,0 +1,79 @@
...
+    fun showDialog() {
+        _activity.let {
+            val markwon = Markwon.builder(it)
+                    .usePlugin(WhatsNewMarkwonPlugin(it.getString(R.string.github_url)))
+                    .usePlugin(LinkifyPlugin.create())
+                    .build()
+            val inflater = _activity.layoutInflater
+            // Pass null as the parent view because its going in the dialog layout.
+            val view = inflater.inflate(R.layout.whats_new, null)
+            val textView = view.findViewById<TextView>(R.id.whatsNewTextView)
```
paetz, line 68: Unnecessary variable. Just pass `findViewById<>()` directly into `setMarkdown()`.

app/src/main/java/com/yacgroup/yacguide/WhatsNew.kt
```diff
@@ -0,0 +1,79 @@
...
+     * Render Markdown code and show dialog.
+     */
+    @SuppressLint("InflateParams")
+    fun showDialog() {
+        _activity.let {
+            val markwon = Markwon.builder(it)
+                    .usePlugin(WhatsNewMarkwonPlugin(it.getString(R.string.github_url)))
+                    .usePlugin(LinkifyPlugin.create())
+                    .build()
+            val inflater = _activity.layoutInflater
```
paetz, line 65: `it` instead of `_activity`

app/src/main/java/com/yacgroup/yacguide/WhatsNew.kt
```diff
@@ -0,0 +1,79 @@
...
+
+class WhatsNew(private var _activity: AppCompatActivity) {
+
+    /*
+     * Read release notes from Markdown files, add an additional heading for each release
+     * and merge them together. Show only a limited number of versions.
+     */
+    private fun _getReleaseNotes(): String {
+        val maxVersionsShow = 10
+        val releaseNotesPath = "release-notes"
```
paetz, line 37: Pls declare this path as a class constant.

app/src/main/res/values/strings.xml
```diff
@@ -135,4 +135,7 @@
         <item>Projektierte Felsen \u2117</item>
     </string-array>
 
+    <!-- What's New dialog -->
+    <string name="whats_new">Was ist neu</string>
```
paetz, line 139: 1. There is already a `dialog` section in this file.
2. Maybe "Neuigkeiten" is the better slogan here?

app/src/main/java/com/yacgroup/yacguide/CountryActivity.kt
```diff
@@ -67,4 +78,26 @@ class CountryActivity : UpdatableTableActivity() {
     override fun getLayoutId(): Int {
         return R.layout.activity_country
     }
+
+    /*
+     * Check whether the "What's New" dialog is shown or not. This is decided by comparing
+     * the version stored in the settings with the current app version.
+     */
+    private fun _init() {
+        val settingsKeyVerCode = "version_code"
```
paetz, line 87: Please declare this setting in `strings.xml`. All other settings are declared there, too.

app/src/main/java/com/yacgroup/yacguide/CountryActivity.kt
```diff
@@ -67,4 +78,26 @@ class CountryActivity : UpdatableTableActivity() {
...
+        var curVerCode = 0
```
paetz, line 88: To keep code clean and secure, try to avoid non-const. variables wherever possible. Simple solution here would be:
```
val currVersionCode = try {
packageManager.getPackageInfo(packageName, 0).versionCode
} catch (e: Exception) {
0
}
```

app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
```diff
@@ -68,4 +72,9 @@ class AboutActivity : BaseNavigationActivity() {
             Toast.makeText(this, R.string.no_webbrowser_available, Toast.LENGTH_SHORT).show()
         }
     }
+
+    @Suppress("UNUSED_PARAMETER")
+    fun showWhatsNew(v: View) {
+        _whatsNew.showDialog()
```
paetz, line 78: It seems that there is no need for `_whatsNewInfo` to be a class member. I would simply do `WhatsNewInfo(this).showDialog()` and remove the member variable.

app/src/main/java/com/yacgroup/yacguide/CountryActivity.kt
```diff
@@ -67,4 +78,26 @@ class CountryActivity : UpdatableTableActivity() {
     override fun getLayoutId(): Int {
         return R.layout.activity_country
     }
+
+    /*
+     * Check whether the "What's New" dialog is shown or not. This is decided by comparing
+     * the version stored in the settings with the current app version.
+     */
+    private fun _init() {
```
paetz, line 86: This method has nothing to do with Country selection. I would move it the the `WhatsNew` class and call sth like `checkForVersionUpdate()`. In the constructor of this activity, you could then do:
```
val whatsNewInfo = WhatsNew(this)
if (whatsNewInfo.checkForVersionUpdate()) {
whatsNewInfo.showDialog()
} 
```

The `SharedPreferences` can be moved to WhatsNew class, too.

paetz (changes requested):

app/src/main/java/com/yacgroup/yacguide/WhatsNew.kt
```diff
@@ -0,0 +1,79 @@
...
+import java.io.File
+
+class WhatsNew(private var _activity: AppCompatActivity) {
+
+    /*
+     * Read release notes from Markdown files, add an additional heading for each release
+     * and merge them together. Show only a limited number of versions.
+     */
+    private fun _getReleaseNotes(): String {
+        val maxVersionsShow = 10
```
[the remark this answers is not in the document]
chrgernoe: Your are right. One version is enought and in case somebody wants see the full history, it is available at GitHub.

app/src/main/java/com/yacgroup/yacguide/CountryActivity.kt
chrgernoe: The `strings.xml` is not a good place because it will be translated, which is not intended. I recommend to created a new resource file which only handles the preference keys.

chrgernoe:
Thanks for the valuable feedback! I made the requested changes and decided to implement the "show more" functionality.

app/src/main/res/values/strings.xml
```diff
@@ -146,4 +147,8 @@
         <item>Projektierte Felsen \u2117</item>
     </string-array>
 
+    <!-- What's New dialog -->
+    <string name="whats_new">Neuigkeiten</string>
+    <string name="whats_new_show_more">Mehr</string>
```
paetz, line 152: - [ ] Remove this. You already have the resource `R.string.more`

paetz (approved):
Please address the 2 minor things before merging

PR #201, opened 2020-12-28
Rework of about view

### LINKED ISSUE #156: Rework layout of about view
Most apps do left align the text. In addition I would implement that all URLs can be opened directly in the external browser.

chrgernoe:
Fixes #156 

* Use default background (white)
* Left align items
* Make items with URLs clickable

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
@@ -38,30 +38,24 @@ class AboutActivity : BaseNavigationActivity() {
         super.onCreate(savedInstanceState)
         this.title = getString(R.string.menu_about, getString(R.string.app_name))
 
-        _createEntry(getString(R.string.data_src), getString(R.string.data_src_url),null)
-        _createEntry(getString(R.string.privacy_policy), null) { _showPrivacyPolicy() }
-        _createEntry(
-                getString(R.string.source_code_and_support),
-                getString(R.string.github_url),
-                null )
-        _createEntry(getString(R.string.license), getString(R.string.license_url),null)
-        _createEntry(getString(R.string.version), _getAppVersion(), null)
+        _createEntry(getString(R.string.data_src), getString(R.string.data_src_url))
+        _createEntry(getString(R.string.privacy_policy), callback = { _showPrivacyPolicy() })
+        _createEntry(getString(R.string.source_code_and_support), getString(R.string.github_url))
+        _createEntry(getString(R.string.license), getString(R.string.license_url))
+        _createEntry(getString(R.string.version), _getAppVersion())
     }
 
     private fun _createEntry(
             title: String,
-            description: String?,
-            callback: ((view: View) -> Unit)?) {
-        val contentAbout: LinearLayout = findViewById(R.id.aboutContent)
-        val entry = layoutInflater.inflate(R.layout.about_entry, null)
-        if (callback == null && URLUtil.isValidUrl(description)) {
-            entry.setOnClickListener { description?.let { url -> _openUrl(url) } }
-        } else {
-            entry.setOnClickListener(callback)
+            description: String = "",
+            callback: (() -> Unit)? = null) {
+        layoutInflater.inflate(R.layout.about_entry, null).let {
+            val contentAbout = findViewById<LinearLayout>(R.id.aboutContent)
+            it.setOnClickListener { callback?.invoke() ?: _openUrl(description) }
+            it.findViewById<TextView>(R.id.aboutEntryTitle).text = title
+            it.findViewById<TextView>(R.id.aboutEntryDescription).text = description
+            contentAbout.addView(it)
         }
-        entry.findViewById<TextView>(R.id.aboutEntryTitle).text = title
-        entry.findViewById<TextView>(R.id.aboutEntryDescription).text = description
-        contentAbout.addView(entry)
     }
 
     private fun _getAppVersion(): String {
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
```diff
@@ -36,12 +38,33 @@ class AboutActivity : BaseNavigationActivity() {
...
 
-        val textViewAppVersion: TextView = findViewById(R.id.textViewAppVersion)
-        textViewAppVersion.text = "Version ${getAppVersion()}"
-        textViewAppVersion.gravity = Gravity.CENTER_HORIZONTAL
+        _createEntry(getString(R.string.data_src), getString(R.string.data_src_url),null)
+        _createEntry(getString(R.string.privacy_policy), null) { _showPrivacyPolicy() }
+        _createEntry(
+                getString(R.string.source_code_and_support),
+                getString(R.string.github_url),
+                null )
```
paetz, line 46: Delete space behind `null`

app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
```diff
@@ -36,12 +38,33 @@ class AboutActivity : BaseNavigationActivity() {
...
+        _createEntry(getString(R.string.license), getString(R.string.license_url),null)
```
paetz, line 47: add space in front of `null`

app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
```diff
@@ -36,12 +38,33 @@ class AboutActivity : BaseNavigationActivity() {
...
+    private fun _createEntry(
+            title: String,
+            description: String?,
+            callback: ((view: View) -> Unit)?) {
+        val contentAbout: LinearLayout = findViewById(R.id.aboutContent)
+        val entry = layoutInflater.inflate(R.layout.about_entry, null)
+        if (callback == null && URLUtil.isValidUrl(description)) {
+            entry.setOnClickListener { description?.let { url -> _openUrl(url) } }
+        } else {
+            entry.setOnClickListener(callback)
```
paetz, line 60: This will be also triggered if `callback == null` and `description` is not a valid url. Might lead to NullPointerException

app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
```diff
@@ -36,12 +38,33 @@ class AboutActivity : BaseNavigationActivity() {
...
     }
 
-    private fun getAppVersion() : String {
+    private fun _createEntry(
+            title: String,
+            description: String?,
+            callback: ((view: View) -> Unit)?) {
+        val contentAbout: LinearLayout = findViewById(R.id.aboutContent)
+        val entry = layoutInflater.inflate(R.layout.about_entry, null)
+        if (callback == null && URLUtil.isValidUrl(description)) {
```
paetz, line 57: I don't like having two parameters that contradict each other (`description` and `callback`). Too many permutations to be checked. 
1. You better enforce the callback parameter (non-null) and pass `_openUrl()` call to the method in case of `description` being a link. 
2. Since you use the value of `description` for every entry (by assignment to TextView), it should be non-null. Better use default-initialization.
```
_createEntry(titleA, { _showPrivacyBlaBlub() })
_createEntry(titleB, { _openUrl(desc) }, desc)
...

private fun _createEntry(title: String, callback: () -> Unit, description: String = "")
```
The check for valid url should be done in `_openUrl()`

app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
```diff
@@ -36,12 +38,33 @@ class AboutActivity : BaseNavigationActivity() {
         super.onCreate(savedInstanceState)
         this.title = getString(R.string.menu_about, getString(R.string.app_name))
 
-        val textViewAppVersion: TextView = findViewById(R.id.textViewAppVersion)
-        textViewAppVersion.text = "Version ${getAppVersion()}"
-        textViewAppVersion.gravity = Gravity.CENTER_HORIZONTAL
+        _createEntry(getString(R.string.data_src), getString(R.string.data_src_url),null)
+        _createEntry(getString(R.string.privacy_policy), null) { _showPrivacyPolicy() }
```
paetz, line 42: Shouldn't `{ _showPrivacyPolicy() }` be part of the method's argument list?

app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
```diff
@@ -36,12 +38,33 @@ class AboutActivity : BaseNavigationActivity() {
...
+                null )
+        _createEntry(getString(R.string.license), getString(R.string.license_url),null)
+        _createEntry(getString(R.string.version), _getAppVersion(), null)
     }
 
-    private fun getAppVersion() : String {
+    private fun _createEntry(
+            title: String,
+            description: String?,
+            callback: ((view: View) -> Unit)?) {
```
paetz, line 54: Why does the callback expect a view as parameter?

paetz (changes requested):

paetz:
- [ ] Before merging, please add a milestone to the issue.

app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
chrgernoe: Why?

app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
chrgernoe: Android Studio suggestions: _Lambda argument should be moved out of parentheses_ . It is a bit strange syntax. There is a another way to get rid of this warning.

app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
chrgernoe: Because `setOnClickListener` expects this. I fixed this.

app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
chrgernoe: I understand your general point and fixed that. 

However, I don't like the solution to  pass the description twice. I found an alternative solution.

paetz (approved):

PR #204, opened 2021-01-22
Fix issue with preserve of date and edited text (fixes #202)

### LINKED ISSUE #202: Opening PartnerActivity does not preserve edited date and notes

chrgernoe:


app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
```diff
@@ -45,63 +47,60 @@ import java.util.Calendar
...
 
-        findViewById<View>(R.id.notesEditText).onFocusChangeListener = View.OnFocusChangeListener { view, _ ->
-            val imm = view.context.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
-            imm.hideSoftInputFromWindow(view.windowToken, 0)
+        findViewById<EditText>(R.id.notesEditText).let {
+            it.onFocusChangeListener = View.OnFocusChangeListener { view, _ ->
+                val imm = view.context.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
+                imm.hideSoftInputFromWindow(view.windowToken, 0)
+            }
+            it.addTextChangedListener(object : TextWatcher {
```
paetz, line 76: This is ok but a little bit like "mit Kanonen auf Spatzen". Why not saving all stuff to `_ascend` when this activity is paused for adding partners (callback `onPause()`)?

app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
```diff
@@ -45,63 +47,60 @@ import java.util.Calendar
...
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_ascend)
 
         _db = DatabaseWrapper(this)
 
-        _ascend = _db.getAscend(intent.getIntExtra(IntentConstants.ASCEND_KEY, DatabaseWrapper.INVALID_ID))
+        // The database query is null, if this activity is called from the route DescriptionActivity.
+        _existingAscend = _db.getAscend(intent.getIntExtra(IntentConstants.ASCEND_KEY,
```
paetz, line 61: It's quite confusing having 2 `Ascend` objects here. Wouldn't it work to just set `_ascend = _db.getAscend(...from_ASCEND_KEY...) ?: Ascend()`? The field `id` is initialized by `0` if we create a new ascend according to file `Ascend.kt`. So instead of checking for `_existingAscend` in method `enter()` you could check for `_ascend.id == 0`. Doing so, we could get rid of this confusing `_existingAscend` variable.

app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
```diff
@@ -45,63 +47,60 @@ import java.util.Calendar
...
+                DatabaseWrapper.INVALID_ID))
+        _ascend = _existingAscend ?: Ascend().apply { routeId = DatabaseWrapper.INVALID_ID }
```
paetz, line 63: I don't understand the body of `apply`? If `_existingAscend` is null, the user has entered via DescriptionActivity and hence, `ROUTE_KEY` must have been set which leads to a valid route id.

paetz (changes requested):

app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
chrgernoe: Thanks for the hint. I just did not know about this feature.

chrgernoe: I like my solution more because otherwise we have two different locations where we have to save the text to the ascend object. First time in the possible `onPause` method and a second time in the `enter` method.

paetz: ok for me. Maybe we can unify both in the future.

paetz (approved):

PR #206, opened 2021-02-08
Fix missing initialization of partner IDs

### LINKED ISSUE #205: App crashes, if new tour book entry is created without visiting the partner selection activity
This issue is caused by patch #204.

chrgernoe:
Fixes #205

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
@@ -59,7 +59,7 @@ class AscendActivity : AppCompatActivity() {
         // The database query is different from null,
         // if this activity is called from the TourbookActivity or TourbookAscendActivity.
         _ascend = _db.getAscend(intent.getIntExtra(IntentConstants.ASCEND_KEY,
-                DatabaseWrapper.INVALID_ID)) ?: Ascend()
+                DatabaseWrapper.INVALID_ID)) ?: Ascend().apply { partnerIds = ArrayList() }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
```diff
@@ -59,7 +59,7 @@ class AscendActivity : AppCompatActivity() {
         // The database query is different from null,
         // if this activity is called from the TourbookActivity or TourbookAscendActivity.
         _ascend = _db.getAscend(intent.getIntExtra(IntentConstants.ASCEND_KEY,
-                DatabaseWrapper.INVALID_ID)) ?: Ascend()
+                DatabaseWrapper.INVALID_ID)) ?: Ascend().apply { partnerIds = ArrayList() }
```
paetz, line 62: The ascend's `notes` field is also initialized with `null` on creation.

- [ ] Check if this is a problem and initialize with `""` here if necessary.

paetz (dismissed):

app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
chrgernoe: It is not a problem because `onCreate` calls `_displayContent` first. This sets the note text where `null` is handled automatically. However, this `null` will be converted to an empty string by the `EditText` object and stored in the ascend object via on-change callback. 

Even, if it is not necessary, I will add the initialization, which is a bit cleaner and easier to understand.

Issue #210, opened 2021-02-14
Implement user-friendly way to report bugs or request features via email


chrgernoe:
A normal app user has no GitHub account to report a bug or request a feature. Therefore a convenient way via email needs to be implemented. 

**Proposal:**
* Add new entry to about view
* Let the user choose between bug, feature and a general contact (merge with #209)
* Create email templates for each concern
* In case of a bug report, automatically extract Android and app version
* Implement possibility to send a bug report, if an import error occurs, including the JSON data which causes the issue (see #211)

PR #211, opened 2021-02-16
Improve error message in case of a tourbook import error

### LINKED ISSUE #208: Tourbook import error for data set > 6000 entries
Jacob T. filed a bug report that tourbook import from sandsteinklettern.de resulted in an error. He had a data et of > 6000 entries.
ToDo:
- Jacob may provide his tourbook for debugging
- report exact JSON or IO error for future bugs

chrgernoe:
Fixes #208 

@paetz: What do you say about this solution?

If there is e.g. a syntax issue in the JSON file, the error message looks e.g. like `Unterminated object at character 42 of <JSON string>`. It would be possible to parse this message, cut only the characters around the mentioned position and so show the exact issue. However, I think this may a bit too over-engineered at this point. Additionally the code may break in case the message changes in the future.

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
@@ -18,7 +18,7 @@
 package com.yacgroup.yacguide
 
 import android.app.Activity
-import android.app.Dialog
+import android.app.AlertDialog
 import android.content.Context
 import android.content.Intent
 import android.content.SharedPreferences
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
```diff
@@ -113,30 +113,41 @@ class TourbookActivity : BaseNavigationActivity() {
...
+            setIcon(android.R.drawable.ic_dialog_alert)
+            setNegativeButton(R.string.cancel) { dialog, _ -> dialog.dismiss()}
+            setPositiveButton(R.string.ok) { dialog, _ ->
+                try {
+                    TourbookExporter(_db, contentResolver).importTourbook(uri)
+                    _makeToast(getString(R.string.tourbook_import_successfull))
+                } catch (e: Exception) {
+                    _showImportError(e.message.toString())
+                }
+                _initYears()
```
paetz, line 128: Isn't there a `dialog.dismiss()` missing here?

app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
```diff
@@ -113,30 +113,41 @@ class TourbookActivity : BaseNavigationActivity() {
...
-        confirmDialog.findViewById<View>(R.id.noButton).setOnClickListener {
-            confirmDialog.dismiss()
-        }
-        confirmDialog.setCanceledOnTouchOutside(false)
-        confirmDialog.setCancelable(false)
         confirmDialog.show()
     }
 
+    private fun _showImportError(errMsg: String) {
+        val dialog = AlertDialog.Builder(this).apply {
```
paetz, line 135: I think, this dialog makes sense. In a follow-up PR (issue #210) I would suggest to add a button here to directly file a bug report with the given error. I would furthermore suggest to also add #210 to milestone v0.3.0 and make the release afterwards.

- [ ] Add this ToDo to #210 description.
- [ ] Set milestone of #210 to v0.3.0

paetz (dismissed):

paetz:
> Fixes #208
> 
> @paetz: What do you say about this solution?
> 
> If there is e.g. a syntax issue in the JSON file, the error message looks e.g. like `Unterminated object at character 42 of <JSON string>`. It would be possible to parse this message, cut only the characters around the mentioned position and so show the exact issue. However, I think this may a bit too over-engineered at this point. Additionally the code may break in case the message changes in the future.

I wouldn't make it too complicated. I think, your solution is ok: Show only a few error lines (maybe 500 characters is even a bit much) and provide the opportunity to send the complete error as a bug report (in #210).

app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
chrgernoe: When the user touches any of the action buttons created with an AlertDialog.Builder, the system dismisses the dialog for you.

PR #217, opened 2021-02-28
Improve app style

chrgernoe:
I tried to make the UI consistent regarding color. I removed most of the hard-coded colors in the layout files and tried to achieve the same with styles which is the recommended way. However, I left most of the UI in original state.

There will be a future PR, which will change all the icons to material design. In this PR, I just changed the icons in the app bar because otherwise I couldn't implement the correct coloring.


paetz:
The table activity background below a table is now white instead of gray.
- [ ] Please revert this.

paetz:
- [ ] Please change the close (cross) button in the comments dialog to new style, too.

paetz (dismissed):
Result looks good. Please address my comments, at least the one regarding table background.

chrgernoe:
>     * Please change the arrow buttons in TourbookAscendActivity, too. (There are arrows for multiple ascends of the same route).

It is already changed because the activity uses the the same layout file?

chrgernoe:
> The table activity background below a table is now white instead of gray.

OK. I thought it might look a bit more friendly, if not everything is grey. Maybe we should in general make all the grey colors a bit more lighter.

paetz:
> > The table activity background below a table is now white instead of gray.
> 
> OK. I thought it might look a bit more friendly, if not everything is grey. Maybe we should in general make all the grey colors a bit more lighter.

paetz:
> > ```
> > * Please change the arrow buttons in TourbookAscendActivity, too. (There are arrows for multiple ascends of the same route).
> > ```
> 
> It is already changed because the activity uses the the same layout file?

You're right, must've overseen it.

paetz (approved):

PR #222, opened 2021-03-21
Added number picker for year selection in tourbook (#179)

### LINKED ISSUE #179: Make year selection in tourbook more user-friendly
Replace the arrows for selecting a year by a year-only date picker

paetz:
Resolves #179

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
@@ -349,6 +349,6 @@ class TourbookActivity : BaseNavigationActivity() {
         }
         Arrays.sort(_availableYears)
 
-        return _availableYears.last()
+        return if (_availableYears.isEmpty()) 0 else _availableYears.last()
     }
 }
--- a/app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
@@ -30,6 +30,7 @@ import android.view.Menu
 import android.view.MenuItem
 import android.view.View
 import android.widget.LinearLayout
+import android.widget.NumberPicker
 import android.widget.TextView
 import android.widget.Toast
 
@@ -82,12 +83,18 @@ class TourbookActivity : BaseNavigationActivity() {
         _db = DatabaseWrapper(this)
         _customSettings = getSharedPreferences(getString(R.string.preferences_filename), Context.MODE_PRIVATE)
 
-        _initYears()
+        _currentYear = _initYears()
+        _displayContent()
     }
 
     override fun onResume() {
         super.onResume()
-        _initYears(resetCurrentYear = false)
+        val mostRecentYear = _initYears()
+        if (!_availableYears.contains(_currentYear)) {
+            // This happens if the current year's only ascend has been removed
+            _currentYear = mostRecentYear
+        }
+        _displayContent()
     }
 
     public override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
@@ -98,6 +105,26 @@ class TourbookActivity : BaseNavigationActivity() {
                 IntentConstants.REQUEST_OPEN_TOURBOOK -> data?.data?.also { uri -> _import(uri) }
                 IntentConstants.REQUEST_SAVE_TOURBOOK -> data?.data?.also { uri -> _export(uri) }
             }
+        }
+    }
+
+    @Suppress("UNUSED_PARAMETER")
+    fun chooseYear(v: View) {
+        val dialog = AlertDialog.Builder(this).apply {
+            setTitle(R.string.select_year)
+            setView(R.layout.numberpicker)
+        }.create()
+        // we need to call show() before findViewById can be used.
+        dialog.show()
+        val yearPicker = dialog.findViewById<NumberPicker>(R.id.yearPicker)!!
+        yearPicker.minValue = 0
+        yearPicker.maxValue = _availableYears.size - 1
+        yearPicker.displayedValues = _availableYears.map { if (it == 0) getString(R.string.undefined) else it.toString() }.toTypedArray()
+        yearPicker.value = _availableYears.indexOf(_currentYear)
+        yearPicker.setOnClickListener {
+            _currentYear = _availableYears[yearPicker.value]
+            _displayContent()
+            dialog.dismiss()
         }
     }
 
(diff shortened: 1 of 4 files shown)


app/src/main/res/layout/numberpicker.xml
```diff
@@ -0,0 +1,25 @@
...
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+  -->
+
```
chrgernoe, line 18:
```suggestion
<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
    android:layout_height="match_parent"
    android:layout_width="match_parent"
    android:gravity="center">
```

app/src/main/res/layout/numberpicker.xml
```diff
@@ -0,0 +1,25 @@
...
+<NumberPicker xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/yearPicker"
+    android:layout_height="wrap_content"
+    android:layout_width="match_parent"
+    android:descendantFocusability="blocksDescendants"/>
+
+
```
chrgernoe, line 25:
```suggestion
</LinearLayout>
```

app/src/main/res/layout/numberpicker.xml
```diff
@@ -0,0 +1,25 @@
...
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+  -->
+
+<NumberPicker xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/yearPicker"
+    android:layout_height="wrap_content"
+    android:layout_width="match_parent"
```
chrgernoe, line 22: Please use `wrap_content` instead and wrap the whole thing into a `LinearLayout`. This makes sure that the number picker looks a bit nicer.

app/src/main/res/layout/numberpicker.xml
```diff
@@ -0,0 +1,25 @@
...
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+  -->
+
+<NumberPicker xmlns:android="http://schemas.android.com/apk/res/android"
```
chrgernoe, line 19: The namespace attribute at this point is no longer required, if it is wrapped by the linear layout.

app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
```diff
@@ -309,17 +340,14 @@ class TourbookActivity : BaseNavigationActivity() {
         startActivityForResult(intent, IntentConstants.REQUEST_SAVE_TOURBOOK)
     }
 
-    private fun _initYears(resetCurrentYear: Boolean = true) {
+    private fun _initYears(): Int {
```
chrgernoe, line 343: I think the method name no longer match the implemented function.

app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
```diff
@@ -205,7 +236,7 @@ class TourbookActivity : BaseNavigationActivity() {
 
         val layout = findViewById<LinearLayout>(R.id.tableLayout)
         layout.removeAllViews()
-        (findViewById<View>(R.id.currentTextView) as TextView).text = if (_currentYear == 0) "" else _currentYear.toString()
+        (findViewById<View>(R.id.currentTextView) as TextView).text = if (_currentYear == 0) getString(R.string.undefined) else _currentYear.toString()
```
chrgernoe, line 239: 1. The id `R.id.currentTextView` is bit confusing.
2. I prefer the previous solution, which showed no year at all in case the year is zero. I think, from the user perspective _undefined_ makes it not clearer.

chrgernoe (changes requested):

app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
paetz: I believe, it's the other way around :) That it didn't match before. However, I'll try to find a better one.

app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
paetz: For the tourbook view maybe, but not for the NumberPicker view. There it would be odd to have an empty field for selecting undefined ascends and I tried to avoid some conversions here like `"" <--> "unbestimmt"`

app/src/main/res/layout/numberpicker.xml
paetz: Your suggestion looks odd on Android 6, e.g. no centering of the number picker. I will go with the current one apart from `wrap_content`.

chrgernoe:
With this code base the app crashes on a fresh installation.

chrgernoe (changes requested):
With this code base the app crashes on a fresh installation.

paetz:
> With this code base the app crashes on a fresh installation, if the user selects the tour book.

Wtf. Which Android version? I observed no problems on Android 6.

paetz:
> With this code base the app crashes on a fresh installation, if the user selects the tour book.

Thx for testing this

app/src/main/res/layout/numberpicker.xml
chrgernoe: What exactly you don't like? It looks the same for each Android version.

paetz: In the Android 6 emulator (API level 23), your suggestion looks the same as mine except that the numbers/years are not centered as in my solution.

app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
paetz: However, I've reverted to empty string in the tourbook view.

chrgernoe (approved):

PR #226, opened 2021-04-06
Fix disappearing partner selection, if search field is used

### LINKED ISSUE #224: Partner selection gets lost, if partner search  field is used
If the user selects one or more partners in the tour book entry dialog and then uses the partner search field, the previous selection gets lost.

chrgernoe:
Fixes #224

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/PartnersActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/PartnersActivity.kt
@@ -147,32 +147,37 @@ class PartnersActivity : AppCompatActivity() {
                 innerLayout.addView(it)
             }
 
-            val editButton = ImageButton(this)
-            editButton.id = View.generateViewId()
-            editButton.setImageResource(android.R.drawable.ic_menu_edit)
-            editButton.setOnClickListener {
-                _updatePartner(partner, getString(R.string.dialog_text_change_partner))
+            val editButton = ImageButton(this).apply {
+                id = View.generateViewId()
+                setImageResource(android.R.drawable.ic_menu_edit)
+                setOnClickListener {
+                    _updatePartner(partner, getString(R.string.dialog_text_change_partner))
+                }
             }
             innerLayout.addView(editButton)
 
-            val deleteButton = ImageButton(this)
-            deleteButton.id = View.generateViewId()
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/yacgroup/yacguide/PartnersActivity.kt
```diff
@@ -207,25 +193,52 @@ class PartnersActivity : AppCompatActivity() {
...
+            view
+        }
+        val builder = AlertDialog.Builder(this).apply {
+            setTitle(dialogTitle)
+            setView(view)
+            setNegativeButton(R.string.cancel) { dialog, _ -> dialog.dismiss()}
+            // Set to null. We override the onclick below.
+            setPositiveButton(R.string.ok, null)
         }
+        // This solution prevents that the dialog is closed automatically,
```
paetz, line 217: Thx for documenting this. It seems that closing the dialog after clicking on negative/positive button is default behavior? If so, why is there a `dialog.dismiss()` in `setNegativeButton`? Same for `deletePartner`dialog

paetz (approved):

app/src/main/java/com/yacgroup/yacguide/PartnersActivity.kt
```diff
@@ -144,56 +133,53 @@ class PartnersActivity : AppCompatActivity() {
...
             val innerLayout = RelativeLayout(this)
             innerLayout.setBackgroundColor(Color.WHITE)
 
-            val checkBox = CheckBox(this)
-            checkBox.text = partnerName
-            if (_selectedPartnerIds.contains(partner.id)) {
-                checkBox.isChecked = true
+            CheckBox(this).let {
+                it.text = partner.name.orEmpty()
+                if (_selectedPartnerIds.contains(partner.id)) {
```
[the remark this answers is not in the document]
chrgernoe: `false` seems to be the default value because I didn't change this particular code line and it worked before, too. However, the suggested solution is safer and needs only one code line.

app/src/main/java/com/yacgroup/yacguide/PartnersActivity.kt
chrgernoe: You are right, We can skip this and just set the listener to `null`.

PR #228, opened 2021-04-15
Fix broken ascend update

### LINKED ISSUE #227: Change of style from e.g. solo to botch generates an invalid database entry
If a tour book entry e.g. with style _Solo_ is added and after saving, the style is changed to e.g. _Gesackt_, the entry is still coloured like a lead ascend (if enabled). It looks like there are two entries in the database because the entry can be deleted but the "eye" symbol is still there and shows an empty entry. The app crashes, if the delete button is pressed in the empty entry.
This doe not happen, if an entry is changed from e.g. _Solo_ to _Nachstieg_.

paetz:
Resolves #227

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
@@ -58,7 +58,7 @@ class AscendActivity : AppCompatActivity() {
         _db = DatabaseWrapper(this)
 
         // The database query is different from null,
-        // if this activity is called from the TourbookActivity or TourbookAscendActivity.
+        // if this activity is called from the TourbookAscendActivity.
         _outdatedAscend = _db.getAscend(intent.getIntExtra(IntentConstants.ASCEND_KEY, DatabaseWrapper.INVALID_ID))
         // The intent constant ROUTE_KEY is only available, if this activity is called from
         // the route DescriptionActivity.
--- a/app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
@@ -48,6 +48,7 @@ class AscendActivity : AppCompatActivity() {
 
     private lateinit var _db: DatabaseWrapper
     private lateinit var _ascend: Ascend
+    private var _outdatedAscend: Ascend? = null
     private var _route: Route? = null
 
     override fun onCreate(savedInstanceState: Bundle?) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
```diff
@@ -58,17 +59,22 @@ class AscendActivity : AppCompatActivity() {
 
         // The database query is different from null,
         // if this activity is called from the TourbookActivity or TourbookAscendActivity.
```
chrgernoe, line 61: This comment is not correct. This activity is only called from `TourbookAscendActivity`. right?

app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
```diff
@@ -58,17 +59,22 @@ class AscendActivity : AppCompatActivity() {
...
+        _ascend = Ascend().apply {
+                    id = _outdatedAscend?.id ?: 0
+                    routeId = _outdatedAscend?.routeId ?: _route!!.id
+                    styleId = _outdatedAscend?.styleId ?: 0
+                    year = _outdatedAscend?.year ?: 0
+                    month = _outdatedAscend?.month ?: 0
+                    day = _outdatedAscend?.day ?: 0
+                    partnerIds = _outdatedAscend?.partnerIds ?: ArrayList()
+                    notes = _outdatedAscend?.notes ?: ""
+                }
```
chrgernoe, line 77: You initialize all properties again, however the `Ascend` class does it already for some of them. I think a more consistent way would be to do it for just the required ones.

What returns `_route!!.id`, if `_route` is `null`?

app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
paetz: Correct but this comment is actually from you :)

app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
paetz: 1. `_route` can only be `null` if `_outdatedAscend` is not `null`. So no problem here: However, I guess it's again time to have some unit tests :)
2. Yes, this initialization of `_ascend` is needed because I want to copy `_outdatedAscend` and there is no `copy` method in Kotlin for non-`data`-classes

app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
chrgernoe: OK, I fixed it.

PR #230, opened 2021-04-19
Feature/193 provide unknown ascend style

### LINKED ISSUE #193: By default select an invalid ascend style in the tourbook UI
At the moment, the default is `Solo`, which is a valid style. Hence, if the user forget to change it, there will be no warning and it may lead to an incorrect entry.

**Proposal:**
By default select an invald style and show e.g. the string `Select style`. If the entry is saved, this shall be checked and an error message shall be shown, if required.

paetz:
Resolves #193


chrgernoe (changes requested):
I think, there are too many location in the code, where you compare against the bit mask zero and two positions where you add the same comment to explain what zero means. Please define a constant with a reasonable name for this. Even, if it feels a bit over-engineered but I think this makes the code more readable.

chrgernoe (approved):

PR #232, opened 2021-04-20
Make all dialogs non-cancelable

### LINKED ISSUE #231: Make all dialogs non-cancelable
The cancelable behaviour was introduced by switching to the `AlertDialog.Builder` class.

chrgernoe:
Fixes #231

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
@@ -25,7 +25,6 @@ import android.webkit.URLUtil
 import android.widget.LinearLayout
 import android.widget.TextView
 import android.widget.Toast
-import androidx.appcompat.app.AlertDialog
 import com.yacgroup.yacguide.utils.ActivityUtils
 import com.yacgroup.yacguide.utils.ContactUtils
 
@@ -82,7 +81,7 @@ class AboutActivity : BaseNavigationActivity() {
 
     private fun _selectContactConcern() {
         val contactInterface = ContactUtils(this)
-        val dialog = AlertDialog.Builder(this).apply {
+        val builder = DialogWidgetBuilder(this).apply {
             setTitle(getString(R.string.contact_concern))
             setItems(R.array.contactConcerns) { _, which ->
                 when (which) {
@@ -90,8 +89,8 @@ class AboutActivity : BaseNavigationActivity() {
                     1 -> contactInterface.sendBugReport()
                 }
             }
-            setNegativeButton(R.string.cancel) { dialog, _ -> dialog.dismiss()}
+            setNegativeButton()
         }
-        dialog.show()
+        builder.show()
     }
 }
--- a/app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
@@ -82,7 +82,7 @@ class AboutActivity : BaseNavigationActivity() {
 
     private fun _selectContactConcern() {
         val contactInterface = ContactUtils(this)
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/yacgroup/yacguide/DialogWidgetBuilder.kt
```diff
@@ -0,0 +1,37 @@
+/*
```
paetz, line 1: I think this file should be placed inside the `utils` directory

app/src/main/java/com/yacgroup/yacguide/DialogWidgetBuilder.kt
```diff
@@ -0,0 +1,37 @@
...
+ */
+class DialogWidgetBuilder(private var _context: Context): AlertDialog.Builder(_context) {
+
+    init {
+        setCancelable(false)
+    }
+
+    fun setNegativeButton() {
+        // Closing the dialog is the default behaviour. Hence we don't implement a listener here.
+        setNegativeButton(_context.getString(R.string.cancel), null)
```
paetz, line 35: Since you explicitly refer to the super class you should use `super.setNegativeButton(...)`.

app/src/main/java/com/yacgroup/yacguide/DialogWidgetBuilder.kt
```diff
@@ -0,0 +1,37 @@
...
+package com.yacgroup.yacguide
+
+import androidx.appcompat.app.AlertDialog
+import android.content.Context
+
+/*
+ * Use this class for each dialog in the app because it sets some reasonable defaults
+ * and avoids repeating code.
+ */
+class DialogWidgetBuilder(private var _context: Context): AlertDialog.Builder(_context) {
```
paetz, line 27: Please add the dialogs title as an argument and call `setTitle()` in the constructor.

app/src/main/java/com/yacgroup/yacguide/DialogWidgetBuilder.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    init {
+        setCancelable(false)
+    }
+
+    fun setNegativeButton() {
```
paetz, line 33: Please add an API like 
```
setPositiveButton(callback) {
    super.setPositiveButton(_context.getString(R.string.ok), callback)
}
```
We always use `R.string.ok` as positive button string.

paetz (changes requested):

app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
```diff
@@ -82,16 +82,15 @@ class AboutActivity : BaseNavigationActivity() {
...
                 when (which) {
                     0 -> contactInterface.sendFeedback()
                     1 -> contactInterface.sendBugReport()
                 }
             }
-            setNegativeButton(R.string.cancel) { dialog, _ -> dialog.dismiss()}
+            setNegativeButton()
         }
-        dialog.show()
+        builder.show()
```
paetz, line 94: I'm a little bit confused. In some cases you call `show()` on a dialog (created by `DialogWidgetBuilder(...).create()`). But here and in some other code spots you call `show()` on the `DialogWidgetBuilder` object itself. Is this intended? If yes, ignore this and just merge.

paetz (approved):
Please check my last comment before merging.

app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
chrgernoe: I went through all the dialog code again and did some further optimization. Normally, I use `show()` of the builder instance. Which I think, is the intended way. However, in some special cases I have to do it a bit different. In these cases, there should be a comment for the reason.

paetz (approved):

PR #238, opened 2021-05-04
Added global project and botch filters

### LINKED ISSUE #186: Add projected rocks search to navigation in UpdatableTableActivity

paetz:
Resolves #186


app/src/main/java/com/yacgroup/yacguide/database/DatabaseWrapper.kt
```diff
@@ -56,18 +56,30 @@ class DatabaseWrapper(context: Context) {
...
 
     fun getRocksForRegion(regionId: Int) = _db.rockDao().getAllInRegion(regionId)
 
+    fun getProjectedRocksForRegion(regionId: Int) = _db.rockDao().getAllInRegionForStyle(regionId, AscendStyle.ePROJECT.id)
+
+    fun getBotchedRocksForRegion(regionId: Int) = _db.rockDao().getAllInRegionForStyle(regionId, AscendStyle.eBOTCHED.id)
+
     fun getRocksForCountry(countryName: String) = _db.rockDao().getAllInCountry(countryName)
 
+    fun getProjectedRocksForCounty(countryName: String) = _db.rockDao().getAllInCountryForStyle(countryName, AscendStyle.ePROJECT.id)
```
paetz, line 73: Typo: `...County`

app/src/main/java/com/yacgroup/yacguide/UpdatableTableActivity.kt
```diff
@@ -110,31 +112,48 @@ abstract class UpdatableTableActivity : TableActivity(), UpdateListener {
...
+    protected abstract fun searchProjects(): List<Rock>
+
+    protected abstract fun searchBotches(): List<Rock>
+
+    private fun _searchRockGlobally(filter: ClimbingObjectFilter = ClimbingObjectFilter.eNone) {
+        when (filter) {
+            ClimbingObjectFilter.eProject -> _goToFilteredRocksView(searchProjects())
+            ClimbingObjectFilter.eBotch -> _goToFilteredRocksView(searchBotches())
+            else -> {
+                val searchDialog = Dialog(this)
```
chrgernoe, line 126: I think we can use our new `DialogWidgetBuilder` here, too, if you make the text of the positive button configurable. If it is not possible fo some reason, please use at least class`AppCompatDialog` from `androidx.appcompat.app` instead to be consistent within the project.

chrgernoe (commented):
In my eyes, this makes the UI a bit inconsistent. On one side we have the solution you implemented for the country and sector view and at the other side the solution with the always visible search and filter field for the rock and route view. I think we should unify both.

I like the solution you implemented now with respect to keep the UI clean but the user needs to click a lot due to the filter sub menu. On the other hand, the solution for the rock and route view makes it easy to use but somehow spams the UI. Maybe a compromise would be good e.g. to avoid the filter sub-menu and use dividers in the menu instead to visually separate the options.

app/src/main/java/com/yacgroup/yacguide/UpdatableTableActivity.kt
chrgernoe: > Isn't the DialogWidgetBuilder restricted to Info + Yes/No action? In that case, I have a more complex dialog containing an EditText field.

Not necessarily, see `WhatsNewInfo` class. 

> Another problem is that AlertDialog is closed by default if clicking on OK. We don't want this here.

OK, this is possible to implement (see `PartnersActivity` class) but at the end it has no advantage over the current solution.

paetz:
> * Most of the icons are from Googles Material Design library, right? I would not add our copyright to these files even if they are free to use. Instead, I suggest to add the URL where you found the icon so that it is documented.

The google icons are from the drawable manager inside Android studio. So no URL needed but I'll remove the copyright.

> * Is the symbol for the projects also from the Material Design library? It seems a bit too big and additionally Android complains that the vector path is too long which is bad for performance.

This is an imported vector graphic. The path for the yac logo is as long, too, so why isn't Android complaining there as well? However, there is a built-in `P` icon (without a circle), I could also use, which has a much shorter path. What do you think?

chrgernoe:
> The google icons are from the drawable manager inside Android studio. So no URL needed but I'll remove the copyright.

You are right.

> This is an imported vector graphic. The path for the yac logo is as long, too, so why isn't Android complaining there as well? However, there is a built-in `P` icon (without a circle), I could also use, which has a much shorter path. What do you think?

I just saw the message. This would be one option but brings some inconsistency because at all other places in the app we are using the `P` with a circle around. Another option, you can try, is to merge the `copyright` symbol with this `P` symbol using some program. However, this merged version should get our own copyright placed on top of the file and maybe a comment where the shapes came from.

paetz:
> I just saw the message. This would be one option but brings some inconsistency because at all other places in the app we are using the `P` with a circle around. Another option, you can try, is to merge the `copyright` symbol with this `P` symbol using some program. However, this merged version should get our own copyright placed on top of the file and maybe a comment where the shapes came from.

Programming your own path / merging copyright with parking symbol is pretty complicated. I leave it as is. I don't actually see any complaint by Android regarding path length of my imported symbol.

chrgernoe (approved):

PR #261, opened 2021-11-23
Update to target sdk version 31

### LINKED ISSUE #213: Update to Android API level 31
After a first trail, this seems to require some code changes.

chrgernoe:
Fixes #213


app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
```diff
@@ -70,11 +71,13 @@ class AboutActivity : BaseNavigationActivity() {
...
-            if (openURL.resolveActivity(packageManager) != null) {
-                startActivity(openURL)
-            } else {
+            try {
+                val intent = Intent(Intent.ACTION_VIEW).apply {
+                    data = Uri.parse(url)
+                    flags = Intent.FLAG_ACTIVITY_NEW_TASK
+                }
+                startActivity(intent)
+            } catch (e: ActivityNotFoundException) {
```
paetz, line 80: Check:
- [ ] This is the only possible exception `startActivity()` may throw.

If not, you might want to catch all exceptions by a general type here.
Same for opening map in RouteActivity

paetz (approved):

app/src/main/java/com/yacgroup/yacguide/AboutActivity.kt
chrgernoe: Yes, it is.

PR #267, opened 2021-12-02
Made activities more flexible to get rid of separate activity displaying search results

### LINKED ISSUE #265: Proper update of selected rock list on resuming
Subtask of #251 

Scenario: Searching projects or botches, selecting one and editing/deleting the corresponding ascent. On resuming to the project/botch list, the corresponding rock is still part of it since the list is not updated on resuming.

Problem: The projected/botched rock ids are transmitted to the SelectedRock view on creation and will hence not be updated on resuming.

Solution: Don't pass a list of rock ids to the activity but pass a search level (country, region, sector, rock), the corresponding id/name and do the database query as part of `displayContent()` depending on the search level. To get rid of code duplications, the `SelectedRockActivity` view should be removed while the filtering would be done directly in `RockActivity`.

paetz:
Resolves #265 

Sorry for this huge PR but I could not split it. 

What has been done:
- To update every activity on resuming (see `TableActivity.onResume()`), it was necessary to move every db get request for current objects to activity's `displayContent()` method. 
- To reduce code duplications, I removed `SelectedRockActivity` and use the normal `RockActivity` to display search/filter results. Doing so, I needed to change the way, we call activities a bit which makes them more flexible regarding current search/filter level.


app/src/main/java/com/yacgroup/yacguide/RockActivity.kt
```diff
@@ -160,6 +148,53 @@ class RockActivity : TableActivityWithOptionsMenu() {
...
+        if (activityLevel.level.value < ClimbingObjectLevel.eRock.value) {
+            val sector = db.getSector(rock.parentId)!!
+            if (activityLevel.level.value < ClimbingObjectLevel.eSector.value) {
+                val region = db.getRegion(sector.parentId)!!
+                sectorInfo = "${region.name} ${getString(R.string.arrow)} "
+            }
+            val sectorNames = ParserUtils.decodeObjectNames(sector.name)
+            sectorInfo += sectorNames.first
+            if (sectorNames.second.isNotEmpty()) {
+                sectorInfo += " / " + sectorNames.second
```
paetz, line 192: use `" / ${sectorNames.second}"`

app/src/main/java/com/yacgroup/yacguide/TableActivity.kt
```diff
@@ -20,21 +20,34 @@ package com.yacgroup.yacguide
...
+
+    lateinit var activityLevel: ClimbingObject
     protected lateinit var db: DatabaseWrapper
     protected lateinit var customSettings: SharedPreferences
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+
+        activityLevel = ClimbingObject(
+            level = ClimbingObjectLevel.fromInt(intent.getIntExtra(IntentConstants.CLIMBING_OBJECT_LEVEL, ClimbingObjectLevel.eCountry.value)),
```
paetz, line 47: Default value should be `ClimbingObjectLevel.eUnknown.value`

chrgernoe (approved):
There are so many code changes which I don't understand in detail but I think you know what you do.

paetz:
1. I wanted to get rid of a separate `SelectedRockActivity` and rather display search results in the common `RockActivity`. This means, the parent activity of `RockActivity` can be `CountryActivity`, `RegionActivity` or `SectorActivity`, depending on the search level (= `activity level`). That's why there is a distinction in `RockActivity` regarding the data to be displayed (search rocks for country/region/sector). 
2. To fix the bug (no update of selected rock list on resuming) I needed to put all database get request into `displayContent()` which is called on resume and NOT (as before) pass the found rock ids to the SelectedRock view.

PR #274, opened 2021-12-06
Show corresponding routes on project or botch filtering

### LINKED ISSUE #255: New view SelectedRoutes
Subtask of #251

- Should (similar to #265) be done within `RouteActivity`
- Project and botch menu items on all levels should point to `RouteActivity` passing the search level, the corresponding id/name and the information about filtering type (project/botch), while the database queries are fired within `displayContent()` depending on these information.

paetz:
Resolves #255 

-> project and botch filters now point to `RouteActivity` to show real projects/botches instead of the according rocks

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/RockActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/RockActivity.kt
@@ -37,15 +37,11 @@ class RockActivity : TableActivityWithOptionsMenu() {
     private var _onlyOfficialSummits: Boolean = false
     private var _rockNamePart: String = ""
     private var _filterName: String = ""
-    private var _filterProjects: Boolean = false
-    private var _filterBotches: Boolean = false
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
         _filterName = intent.getStringExtra(IntentConstants.FILTER_NAME).orEmpty()
-        _filterProjects = intent.getBooleanExtra(IntentConstants.FILTER_PROJECTS, false)
-        _filterBotches = intent.getBooleanExtra(IntentConstants.FILTER_BOTCHES, false)
 
         properties = arrayListOf(RockSearchable(this), AscentFilterable(this))
 
@@ -141,6 +137,9 @@ class RockActivity : TableActivityWithOptionsMenu() {
             )
             layout.addView(WidgetUtils.createHorizontalLine(this, 1))
         }
+        if (rocks.isEmpty()) {
+            Toast.makeText(this, R.string.no_rocks_available, Toast.LENGTH_SHORT).show()
+        }
     }
 
     override fun onStop() {
@@ -150,26 +149,10 @@ class RockActivity : TableActivityWithOptionsMenu() {
 
     private fun _getAndFilterRocks(): List<Rock> {
         val rocks = when (activityLevel.level) {
-            ClimbingObjectLevel.eCountry -> {
-                if (_filterProjects) db.getProjectedRocks()
-                else if (_filterBotches) db.getBotchedRocks()
-                else db.getRocks()
-            }
-            ClimbingObjectLevel.eRegion -> {
-                if (_filterProjects) db.getProjectedRocksForCountry(activityLevel.parentName)
-                else if (_filterBotches) db.getBotchedRocksForCountry(activityLevel.parentName)
-                else db.getRocksForCountry(activityLevel.parentName)
-            }
-            ClimbingObjectLevel.eSector -> {
-                if (_filterProjects) db.getProjectedRocksForRegion(activityLevel.parentId)
-                else if (_filterBotches) db.getBotchedRocksForRegion(activityLevel.parentId)
-                else db.getRocksForRegion(activityLevel.parentId)
-            }
-            ClimbingObjectLevel.eRock -> {
-                if (_filterProjects) db.getProjectedRocksForSector(activityLevel.parentId)
-                else if (_filterBotches) db.getBotchedRocksForSector(activityLevel.parentId)
-                else db.getRocksForSector(activityLevel.parentId)
-            }
+            ClimbingObjectLevel.eCountry -> db.getRocks()
+            ClimbingObjectLevel.eRegion -> db.getRocksForCountry(activityLevel.parentName)
+            ClimbingObjectLevel.eSector -> db.getRocksForRegion(activityLevel.parentId)
+            ClimbingObjectLevel.eRock -> db.getRocksForSector(activityLevel.parentId)
             else -> emptyList()
         }
         return if (_filterName.isNotEmpty())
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/yacgroup/yacguide/database/DatabaseWrapper.kt
```diff
@@ -58,37 +58,45 @@ class DatabaseWrapper(context: Context) {
...
+    fun getProjectedRoutesForSector(sectorId: Int) = _db.routeDao().getAllInSectorForStyle(sectorId, AscendStyle.ePROJECT.id)
 
-    fun getBotchedRocks() = _db.rockDao().getAllForStyle(AscendStyle.eBOTCHED.id)
+    fun getBotchedRoutesForSector(sectorId: Int) = _db.routeDao().getAllInSectorForStyle(sectorId, AscendStyle.eBOTCHED.id)
 
-    fun getRock(rockId: Int) =_db.rockDao().getRock(rockId)
+    fun getRoutesForRegion(regionId: Int) = _db.routeDao().getAllInRegion(regionId)
 
-    fun getRockComments(rockId: Int) = _db.rockCommentDao().getAll(rockId)
+    fun getProjectedRoutesForRegion(regionId: Int) = _db.routeDao().getAllInRegionForStyle(regionId, AscendStyle.ePROJECT.id)
```
paetz, line 85: Maybe `getProjectsForRegion()` etc. is more feasible

app/src/main/java/com/yacgroup/yacguide/RockActivity.kt
```diff
@@ -37,15 +37,11 @@ class RockActivity : TableActivityWithOptionsMenu() {
     private var _onlyOfficialSummits: Boolean = false
     private var _rockNamePart: String = ""
     private var _filterName: String = ""
-    private var _filterProjects: Boolean = false
```
paetz, line 40: - [ ] Check this

paetz, line 40: Should be `true` by default

paetz:
> * If a search for a rock and there is no result, for some reason the toast message is shown twice.
Gonna look at this.

> * Is there a way to get rid of the toast message that no rock was found, if I resume to the app. Right now I get the message with every resume. Correct, in my eyes would be that this message is only show after the user hit the search button.
Not really, since `displayContent()` is called on resuming by default. I could at most introduce a boolean value like `dontShowEmptyListHint` as a parameter of this method which is quite nasty.

> * Why is there still the rock search field in the rock view?
I know about this redundancy. But on the rock view it's much more convenient to use the searchbar for searching. However, rock view must remain a `RockSearchable` because we will add the `relevance` filter to this in the next steps.

chrgernoe:
> Gonna look at this. EDIT: Cannot reproduce. But we have a similar issue when searching via searchbar. In this case the hint is shown for every(!) typed letter and the fired toasts seem to be queued to display them all one after the other.

Ok. I have Android 12. Maybe there is a bug in the libraries but I wouldn't just ignore it.

> I think, we could also skip the "no-rocks-found" hint toast. What do you think? This would resolve your 1st and 2nd point.

Then the first issue would just not visible anymore, I think.
Regarding the second issue: Would it not possible to introduce a boolean parameter like `showToast` which is set to `false` in the `onResume()` method? I wouldn't remove this message.

> I know about this redundancy. But on the rock view it's much more convenient to use the searchbar for searching. However, rock view must remain a RockSearchable because we will add the relevance filter to this in the next steps.

Why not just remove the search entry from the menu. With your flexible approach this should be easy?

paetz:
> Then the first issue would just not visible anymore, I think. Regarding the second issue: Would it not possible to introduce a boolean parameter like `showToast` which is set to `false` in the `onResume()` method? I wouldn't remove this message.

Of course, that's what I proposed. But as I said, this means that if you search a rock via searchbar, every typed letter leads to the toast call. These calls are stacked and called one after the other. Result: You type 10 characters (for which no rock is found) and will get the toast displayed 10 times longer than for one single call (because the calls are stacked).
One solution would be to introduce a text field directly in the activity. How about this? Maybe in the middle of the search bar if possible?

> > I know about this redundancy. But on the rock view it's much more convenient to use the searchbar for searching. However, rock view must remain a RockSearchable because we will add the relevance filter to this in the next steps.
> 
> Why not just remove the search entry from the menu. With your flexible approach this should be easy?

As I stated, this is not applicable because we will soon add the `relevance` filter to the rock search dialog that opens via menu click:
```
|-------------------------|
| Name: ______________  |
| Bedeutung >= ________ |
|         <--Search-->          |
|-------------------------|

chrgernoe:
> One solution would be to introduce a text field directly in the activity. How about this? Maybe in the middle of the search bar if possible?

I wouldn't introduce another text field there. It makes this bar even more growded. Can we change the text color of the typed text, if no result was found?

> As I stated, this is not applicable because we will soon add the relevance filter to the rock search dialog that opens via menu click:

Now I understand and I'm fine with this.

chrgernoe:
> > > One solution would be to introduce a text field directly in the activity. How about this? Maybe in the middle of the search bar if possible?
> > 
> > 
> > I wouldn't introduce another text field there. It makes this bar even more crowded. Can we change the text color of the typed text, if no result was found?
> 
> Surely possible (but only in the searchbar, NOT in the search dialog) but this does not solve the toast problem because updating the searchbar search field calls `displayContent()` as well, just like searching via dialog.

What is about two different solutions.

* For the search bar we are using the colored text or maybe show a exclamation mark behind the search field.
* In the search dialog, we change the behavior as followed. If there is no match, we just do not go back to the table activity but show a message or something similar in the search dialog. It's anyway a not so nice behavior that the search dialog closes, even if there is not match. That means for the next trail the user have to open dialog again and type in the text again e.g. in case he just had a minor typo in it.

paetz:
> What is about two different solutions.
> 
> * For the search bar we are using the colored text or maybe show a exclamation mark behind the search field.
> * In the search dialog, we change the behavior as followed. If there is no match, we just do not go back to the table activity but show a message or something similar in the search dialog. It's anyway a not so nice behavior that the search dialog closes, even if there is not match. That means for the next trail the user have to open dialog again and type in the text again e.g. in case he just had a minor typo in it.

I already know what you meant. And that's what I mean with "impossible". Why? The search dialog does not search anything. It just passes a parameter to `RockActivity` what to be searched within `displayContent()`. So the dialog is already closed, when the searching (db request) is actually executed.

chrgernoe:
> > What is about two different solutions.
> > 
> > * For the search bar we are using the colored text or maybe show a exclamation mark behind the search field.
> > * In the search dialog, we change the behavior as followed. If there is no match, we just do not go back to the table activity but show a message or something similar in the search dialog. It's anyway a not so nice behavior that the search dialog closes, even if there is not match. That means for the next trail the user have to open dialog again and type in the text again e.g. in case he just had a minor typo in it.
> 
> Yes, that's what I mean with "impossible". Why? The search dialog does not search anything. It just passes a parameter to `RockActivity` what to be searched within `displayContent()`. (We cannot search within the search dialog and pass rock ids because otherwise `RockActivity` cannot be updated on resume, see #265) So the dialog is already closed, when the searching (db request) is actually executed (within `RockActivity`'s `displayContent()`).

OK. Then let's skip the toast message completely.

chrgernoe (approved):

PR #275, opened 2021-12-15
Use GitHub Actions instead of Travis CI

### LINKED ISSUE #263: Use GitHub Actions instead of Travis CI
Travis CI is no longer supports free usage for open-source projects automatically. There is a way to get CPU credits by requesting them via email but it feels that they don't really want this anymore. (See https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing)

GitHub Actions seems to be a good alternative and has also the advantage that everything is handled in one place. The free plan is 2000 CPU minutes per month which should be enough for the project.

chrgernoe:
Fix #263


paetz (approved):
Great that this is done!

PR #279, opened 2021-12-20
Accelerated rock search - PLEASE UPDATE YOUR SECTORS

### LINKED ISSUE #276: Accelerate rock search
Full-text rock search may be accelerated by factor 10 if filtering is done in sql directly instead of filtering in Kotlin after the database request.

Drawback: We need to change the separator for german and czech names from `\u0000` to something else. I suggest the control sequence `\u001A` (= substitute control sequence).

Consequence for users: If a user gets this version update, sectors or rocks with german and czech names may not be decoded correctly. However, the user can fix this by simply synchronizing the sector again, so I think, it would be ok.

Worst case (the user does not update local database after version update): 
- sector and rock names are displayed with an unknown symbol and not perfectly formatted
- rock search will not work for 2nd name, e.g.: If name is `Turm X\0000Věž X`, the sql filtering may not find this object if searching `Věž`.

@chrgernoe Is it possible to popup a hint for the user to update his/her local database if he/she opens the app right after a version update (like the What'sNew dialog)?

paetz:
Resolves #276

### DIFF
--- a/docs/release-notes/v0.5.0.md
+++ b/docs/release-notes/v0.5.0.md
@@ -1,7 +1,9 @@
 ## Important notice
 
-* To ensure correct display of rocks/routes and a working rock/route search,
-  please update your local database via "Aktualisieren" for each sector.
+* Es müssen alle bereits heruntergeladenen Gebiete via Menüpunkt
+  "Aktualisieren" neu geladen werden. Andernfalls ist eine
+  funktionierende Volltextsuche, sowie die korrekte Anzeige
+  von Felsen/Wegen nicht gewährleistet.
 
 ## What's Changed
 
--- a/docs/release-notes/v0.5.0.md
+++ b/docs/release-notes/v0.5.0.md
@@ -1,0 +1,12 @@
+## Important notice
+
+* To ensure correct display of rocks/routes and a working rock/route search,
+  please update your local database via "Aktualisieren" for each sector.
+
+## What's Changed
+
+
+
+## Bugfixes
+
+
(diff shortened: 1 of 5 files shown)


chrgernoe (changes requested):
Why is the Unicode character fast? Please add a code comment.

chrgernoe:
I tried the rock search before the patch. I don't see a speed issue? What are the conditions?

paetz:
> Why is the Unicode character fast? Please add a code comment.

What do yo mean by "fast"? \u0000 is interpreted as some error case in sql, hence every name part behind this marker can not be detected by sql command `LIKE`

paetz:
> I tried the rock search before the patch. I don't see a speed issue? What are the conditions?

I did some timing measurements:
If only Sächsische and Böhmische Schweiz are locally available, rock search takes approx. 
- 100-200ms with the old approach
- ~10ms with the new approach

docs/release-notes/v0.5.0.md
```diff
@@ -0,0 +1,12 @@
+## Important notice
+
+* To ensure correct display of rocks/routes and a working rock/route search,
```
paetz, line 3: Should be in German, I guess

chrgernoe: Normally we do all the release notes in English but this case is a bit special. German is also fine.

chrgernoe:
> > Why is the Unicode character fast? Please add a code comment.
> 
> What do yo mean by "fast"? "\u0000" (the old/current separator between german and czech name in our db) is interpreted as some error case in sql, hence every name part behind this marker can not be detected by sql command `LIKE`

Understand. Please at this as comment to the code.

chrgernoe (approved):

PR #280, opened 2021-12-20
Highlight downloaded regions

### LINKED ISSUE #278: Add icon to region to indicate that it is locally available
Sometimes, it makes sense to know which regions you have already downloaded (e.g. if you're instructed to update your local database after tourbook import or version update).

Add a proper icon/symbol in front of the region name or color it accordingly to indicate this.

paetz:
Resolves #278

### DIFF
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -38,8 +38,8 @@
     <color name="colorOnSecondaryLight">@color/colorSecondaryDark</color>
     <color name="colorOnSecondaryDark">@color/white</color>
 
-    <color name="colorDownloaded">#ddeeff</color>
     <color name="colorAccent">@color/colorPrimaryLight</color>
+    <color name="colorAccentLight">#e8eaf6</color>
     <color name="colorError">#dd2c00</color>
 
     <color name="colorPrimaryText">@color/colorBlack</color>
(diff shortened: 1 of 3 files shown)


chrgernoe (changes requested):
* I would not add the tick in front of the downloaded regions. In my eyes this makes the layout a bit nervous. My suggestions would be to add an empty box in front of each item but mark it as check, if downloaded. 
* Please don't introduce a "completely" new color. There is a link in the color definition file that describes the color scheme which is used in the app. Please keep with this color either within the primary or secondary color "space".

app/src/main/res/values/colors.xml
```diff
@@ -38,6 +38,7 @@
     <color name="colorOnSecondaryLight">@color/colorSecondaryDark</color>
     <color name="colorOnSecondaryDark">@color/white</color>
 
+    <color name="colorDownloaded">#ddeeff</color>
```
chrgernoe, line 41: Use a more general name so that this color can also be used on other places in the app.

paetz:
> * Please don't introduce a "completely" new color. There is a link in the color definition file that describes the color scheme which is used in the app. Please keep with this color either within the primary or secondary color "space".

Hm, I tried the secondary color, but this is a bit confusing since grey color means "prohibited" in the other views. I'll choose a light color from our "Indigo" color scheme and define it as a new color, I hope that's ok.

chrgernoe (approved):

Issue #120, opened 2020-03-18
Download of all database items required by the tourbook after an import


chrgernoe:
If you make a fresh installation of the app and import an existing tourbook all entries are shown with question marks.
There should be a function to download necessary database items so that all  tourbook entries are correctly shown.

paetz:
This might become a difficult topic. Since you only have route ids, you would need to scan all routes from all sectors. This then means, you'll need to download all sectors of the whole database to check.

chrgernoe:
You are right. Maybe a simple solution would be to store the sector ID in the tour book as well?

paetz:
Yes, this would be the solution but how do you ensure backwards-compatibility in that case?

chrgernoe:
I think, at the moment this not critical because we have no official release. Nevertheless, we can add a piece of code that checks that for each tour book entry a summit ID is associated. If not, just update the entry, if the database entry is available.
This code can be run with every tour book look-up in the app and should be run for the complete tour book, if it is exported. If you import an older version just ignore the missing sector ID and add the sector ID, if the database entry is available.
In general, maybe we should add a version to the exported tour book, so that possible future changes can be detected easily. Similar to the database version.

PR #282, opened 2021-12-21
Possibility to update whole local database

### LINKED ISSUE #281: Region manager for updating/deleting regions
It would be quite handy to have a button in the settings menu to update all locally available sectors at once.

Especially regarding #276, this feature would reduce user's effort to update all o her/his sectors.

paetz:
Resolves #281 

These changes are also the base for updating the database on tourbook import (#120) or even if we decide to make automatic nightly database updates in the future.

Note: Most of the code in `UpdateHandler.kt` is copied from `Updatable.kt`.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -69,6 +69,7 @@
     <string name="dialog_text_change_partner">Kletterpartner ändern</string>
     <string name="dialog_question_delete">Diesen Bereich löschen?</string>
     <string name="dialog_question_delete_ascend">Diese Begehung löschen?</string>
+    <string name="dialog_question_update">Dies aktualisiert alle Gebiete und kann daher einige Zeit dauern.</string>
     <string name="dialog_loading">Lädt...</string>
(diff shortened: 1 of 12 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -69,6 +69,7 @@
     <string name="dialog_text_change_partner">Kletterpartner ändern</string>
     <string name="dialog_question_delete">Diesen Bereich löschen?</string>
     <string name="dialog_question_delete_ascend">Diese Begehung löschen?</string>
+    <string name="dialog_question_update">Dies aktualisiert alle Gebiete und kann daher einige Zeit dauern.</string>
```
chrgernoe, line 72: The text is a bit long. At least ob my screen the text is truncated. Either shorten the text or reduce the text size. However, the latter solution leads to an inconsistent layout.

chrgernoe (changes requested):
I think the preferences activity is not the right place for the update button. I wouldn't expect the button there. Is it possible to integrate the function into the country menu?

PR #284, opened 2021-12-29
Pretty print for tour book export

### LINKED ISSUE #252: Pretty-print for tourbook export
When exporting the tourbook, the user should be able to decide between rough ID-based export (for re-import on a different device e.g.) and pretty-print export (like a name-based .csv file that may not be imported again).

chrgernoe:
Fixes #252

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
@@ -32,6 +32,7 @@ import android.widget.LinearLayout
 import android.widget.NumberPicker
 import android.widget.TextView
 import android.widget.Toast
+import androidx.activity.result.contract.ActivityResultContracts
 
 import com.yacgroup.yacguide.database.*
 import com.yacgroup.yacguide.utils.*
@@ -48,6 +49,19 @@ class TourbookActivity : BaseNavigationActivity() {
     private lateinit var _availableYears: IntArray
     private var _currentYear: Int = 0
     private var _tourbookType: TourbookType = TourbookType.eAscends
+    // Activity result handler
+    private val _export = registerForActivityResult(
+        ActivityResultContracts.StartActivityForResult()){
+        if (it.resultCode == Activity.RESULT_OK) {
+            it.data?.data?.also { uri -> _doExport(uri) }
+        }
+    }
+    private val _import = registerForActivityResult(
+        ActivityResultContracts.StartActivityForResult()){
+        if (it.resultCode == Activity.RESULT_OK) {
+            it.data?.data?.also { uri -> _doImport(uri) }
+        }
+    }
 
     private lateinit var _customSettings: SharedPreferences
 
@@ -97,11 +111,6 @@ class TourbookActivity : BaseNavigationActivity() {
         super.onActivityResult(requestCode, resultCode, data)
         if (resultCode == IntentConstants.RESULT_UPDATED) {
             Toast.makeText(this, R.string.ascend_deleted, Toast.LENGTH_SHORT).show()
-        } else if (resultCode == Activity.RESULT_OK) {
-            when (requestCode) {
-                IntentConstants.REQUEST_OPEN_TOURBOOK -> data?.data?.also { uri -> _import(uri) }
-                IntentConstants.REQUEST_SAVE_TOURBOOK -> data?.data?.also { uri -> _export(uri) }
-            }
         }
     }
 
@@ -137,7 +146,7 @@ class TourbookActivity : BaseNavigationActivity() {
         _displayContent()
     }
 
-    private fun _import(uri: Uri) {
+    private fun _doImport(uri: Uri) {
         DialogWidgetBuilder(this, R.string.warning).apply {
             setMessage(getString(R.string.override_tourbook))
             setIcon(android.R.drawable.ic_dialog_alert)
@@ -183,7 +192,7 @@ class TourbookActivity : BaseNavigationActivity() {
         }.show()
     }
 
-    private fun _export(uri: Uri) {
+    private fun _doExport(uri: Uri) {
         try {
             TourbookExporter(_db, contentResolver).exportTourbook(uri)
             Toast.makeText(this, getString(R.string.tourbook_export_successfull),
@@ -313,7 +322,7 @@ class TourbookActivity : BaseNavigationActivity() {
             addCategory(Intent.CATEGORY_OPENABLE)
             type = "*/*"
         }
-        startActivityForResult(intent, IntentConstants.REQUEST_OPEN_TOURBOOK)
+        _import.launch(intent)
     }
 
     private fun _selectFileExport() {
@@ -321,7 +330,7 @@ class TourbookActivity : BaseNavigationActivity() {
             addCategory(Intent.CATEGORY_OPENABLE)
             type = "*/*"
         }
-        startActivityForResult(intent, IntentConstants.REQUEST_SAVE_TOURBOOK)
+        _export.launch(intent)
     }
 
     private fun _initYears(): Int {
--- a/app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
@@ -51,17 +51,16 @@ class TourbookActivity : BaseNavigationActivity() {
     private lateinit var _tourbookExporter: TourbookExporter
     private var _currentYear: Int = 0
     private var _tourbookType: TourbookType = TourbookType.eAscends
-    // Activity result handler
-    private val _export = registerForActivityResult(
+    private val _exportResultLauncher = registerForActivityResult(
         ActivityResultContracts.StartActivityForResult()){
         if (it.resultCode == Activity.RESULT_OK) {
-            it.data?.data?.also { uri -> _doExport(uri) }
+            it.data?.data?.also { uri -> _export(uri) }
         }
     }
-    private val _import = registerForActivityResult(
+    private val _importResultLauncher = registerForActivityResult(
         ActivityResultContracts.StartActivityForResult()){
         if (it.resultCode == Activity.RESULT_OK) {
-            it.data?.data?.also { uri -> _doImport(uri) }
+            it.data?.data?.also { uri -> _import(uri) }
         }
     }
 
@@ -71,6 +70,9 @@ class TourbookActivity : BaseNavigationActivity() {
         eAscends,
         eBotches
     }
+
+    private val MIME_TYPE_JSON = "application/json"
+    private val MIME_TYPE_CSV = "text/comma-separated-values"
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
@@ -150,7 +152,7 @@ class TourbookActivity : BaseNavigationActivity() {
         _displayContent()
     }
 
-    private fun _doImport(uri: Uri) {
+    private fun _import(uri: Uri) {
         DialogWidgetBuilder(this, R.string.warning).apply {
             setMessage(getString(R.string.override_tourbook))
             setIcon(android.R.drawable.ic_dialog_alert)
@@ -196,7 +198,7 @@ class TourbookActivity : BaseNavigationActivity() {
         }.show()
     }
 
-    private fun _doExport(uri: Uri) {
+    private fun _export(uri: Uri) {
         try {
             _tourbookExporter.exportTourbook(uri)
             Toast.makeText(this, getString(R.string.tourbook_export_successfull),
@@ -324,23 +326,20 @@ class TourbookActivity : BaseNavigationActivity() {
     private fun _selectFileImport() {
         val intent = Intent(Intent.ACTION_OPEN_DOCUMENT).apply {
             addCategory(Intent.CATEGORY_OPENABLE)
-            type = "application/json"
+            type = MIME_TYPE_JSON
         }
-        _import.launch(intent)
+        _importResultLauncher.launch(intent)
     }
 
-    /*
-     * Open dialog to let the user choose the export format.
-     */
     private fun _selectExportFormat() {
-        var mimeType = "application/json"
+        var mimeType = MIME_TYPE_JSON
         DialogWidgetBuilder(this, R.string.export_format).apply {
             setSingleChoiceItems(R.array.exportFormats, 0) {_, which ->
                 when (which) {
                     0 -> _tourbookExporter.exportFormat = TourbookExporter.ExportFormat.eJSON
-                    1 -> _tourbookExporter.exportFormat = TourbookExporter.ExportFormat.eJSONVEROBSE
+                    1 -> _tourbookExporter.exportFormat = TourbookExporter.ExportFormat.eJSONVERBOSE
                     2 -> {
-                        mimeType = "text/comma-separated-values"
+                        mimeType = MIME_TYPE_CSV
                         _tourbookExporter.exportFormat = TourbookExporter.ExportFormat.eCSV
                     }
                 }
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
```diff
@@ -311,17 +324,38 @@ class TourbookActivity : BaseNavigationActivity() {
     private fun _selectFileImport() {
         val intent = Intent(Intent.ACTION_OPEN_DOCUMENT).apply {
             addCategory(Intent.CATEGORY_OPENABLE)
-            type = "*/*"
+            type = "application/json"
```
paetz, line 327: Please declare constant strings for both mime types.

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookExporter.kt
```diff
@@ -41,11 +49,57 @@ class TourbookExporter(
     private val _partnersKey = "partners"
     private val _notesKey = "notes"
 
+    enum class ExportFormat {
+        eJSON,
+        eJSONVEROBSE,
+        eCSV;
```
paetz, line 55: Unnecessary `;`

app/src/main/res/values/strings.xml
```diff
@@ -124,6 +124,7 @@
     <string name="tourbook_import_successfull">Tourenbuch erfolgreich importiert</string>
     <string name="tourbook_import_error">Fehler beim Import</string>
     <string name="tourbook_export_error">Fehler beim Export des Tourenbuchs</string>
+    <string name="export_format">Wähle das Exportformat</string>
```
paetz, line 127: Better: `Exportformat wählen`

app/src/main/res/values/strings.xml
```diff
@@ -171,6 +172,13 @@
         <item>@string/report_error</item>
     </string-array>
 
+    <!-- IMPORTANT: This list must be in sync with the code in the TourbookActivity class. -->
+    <string-array name="exportFormats">
+        <item>JSON (für Export/Import)</item>
```
paetz, line 177:
```
JSON (ID-basiert für Re-Import)
JSON (Namen-basiert)
CSV (Namen-basiert)
```

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookExporter.kt
```diff
@@ -41,11 +49,57 @@ class TourbookExporter(
...
+            _writeStrToUri(uri, writer.toString())
+        }
+    }
+
+    @Throws(IOException::class)
+    private fun _exportAsJson(uri: Uri) {
         val jsonAscends = JSONArray()
-        _db.getAscends().map { jsonAscends.put(ascend2Json(it)) }
-        _writeStrToUri(uri, jsonAscends.toString())
+        _db.getAscends().map {
```
paetz, line 93: Please use `forEach()`. `forEach()` does not create an output list while `map()` does.

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookEntryVerbose.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    init {
+        val route = _db.getRoute(_ascend.routeId)
+        val rock = route?.parentId?.let { it -> _db.getRock(it) }
+        val sector = rock?.parentId?.let { it -> _db.getSector(it) }
+        val region = sector?.parentId?.let { it -> _db.getRegion(it) }
+
+        country = region?.country.orEmpty()
+        regionName = region?.name.orEmpty()
+        sectorName = ParserUtils.decodeObjectNames(sector?.name).first
```
paetz, line 62: Why only exporting german name? Same for rock and route. Should be encoded to sth like `firstName / secondName`

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookEntryVerbose.kt
```diff
@@ -0,0 +1,95 @@
...
+    var rockName: String
+    var routeName: String
+    var notes: String
+    var date: String
+    var style: String
+    var partners: String
+
+    init {
+        val route = _db.getRoute(_ascend.routeId)
+        val rock = route?.parentId?.let { it -> _db.getRock(it) }
```
paetz, line 56: Do you really need to write `it -> _db.getRock(it)`? Isn't `_db.getRock(it)` sufficient?

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookEntryVerbose.kt
```diff
@@ -0,0 +1,95 @@
...
+                TourbookEntryVerbose::class.memberProperties.forEach{
+                    if (!it.name.startsWith("_")) {
+                        add(it.name)
+                    }
+                }
+            }
+        }
+    }
+
+    var country: String
```
paetz, line 44: `lateinit`?

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookEntryVerbose.kt
```diff
@@ -0,0 +1,95 @@
...
+        sectorName = ParserUtils.decodeObjectNames(sector?.name).first
+        rockName = ParserUtils.decodeObjectNames(rock?.name).first
+        routeName = ParserUtils.decodeObjectNames(route?.name).first
+        notes = _ascend.notes.orEmpty()
+        date = "%02d.%02d.%4d".format(_ascend.day, _ascend.month, _ascend.year)
+        style = AscendStyle.fromId(_ascend.styleId)?.styleName.orEmpty()
+        partners = _db.getPartnerNames(_ascend.partnerIds.orEmpty()).joinToString(",")
+    }
+
+    /*
```
paetz, line 71: remove comment. It's clear by name

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookEntryVerbose.kt
```diff
@@ -0,0 +1,95 @@
...
+        return buildMap {
+            TourbookEntryVerbose::class.memberProperties.forEach {
+                if (it.name in keys()) {
+                    put(it.name, it.get(thisObj).toString())
+                }
+            }
+        }
+    }
+
+    /*
```
paetz, line 85: Remove comment. clear by name

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookEntryVerbose.kt
```diff
@@ -0,0 +1,95 @@
...
+     * Return tour book values in the order of the keys
+     */
+    fun values(): List<String> {
+        return buildList {
+            keys().forEach{
```
paetz, line 90: Can't you use `map()` instead of `forEach()`? `map()` automatically creates an output list

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookEntryVerbose.kt
```diff
@@ -0,0 +1,95 @@
...
+import kotlin.reflect.full.memberProperties
+
+class TourbookEntryVerbose(_ascend: Ascend, val _db: DatabaseWrapper) {
+    companion object {
+        /*
+         * Return a list of all property names of this class
+         * which are treated as fields to be exported.
+         */
+        fun keys(): List<String> {
+            return buildList {
```
paetz, line 34: Since you don't have private members, this could be shortened to
`return TourbookEntryVerbose::class.memberProperties.map{ it.name }`

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookEntryVerbose.kt
```diff
@@ -0,0 +1,95 @@
...
+        date = "%02d.%02d.%4d".format(_ascend.day, _ascend.month, _ascend.year)
+        style = AscendStyle.fromId(_ascend.styleId)?.styleName.orEmpty()
+        partners = _db.getPartnerNames(_ascend.partnerIds.orEmpty()).joinToString(",")
+    }
+
+    /*
+     * Return tour book entry as map
+     */
+    fun asMap(): Map<String, String> {
+        val thisObj = this
```
paetz, line 75: I'd prefer `val self = this`

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookEntryVerbose.kt
paetz: Can `keys()` actually be private?

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookExporter.kt
```diff
@@ -41,11 +49,57 @@ class TourbookExporter(
     private val _partnersKey = "partners"
     private val _notesKey = "notes"
 
+    enum class ExportFormat {
+        eJSON,
+        eJSONVEROBSE,
```
paetz, line 54: Typo: `VEROBSE`

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookExporter.kt
```diff
@@ -41,11 +49,57 @@ class TourbookExporter(
...
+        eCSV;
+    }
+
+    var exportFormat: ExportFormat = ExportFormat.eJSON
```
paetz, line 58: Dangerous without unit tests. Public state variables are pretty error-prone. This info should actually be passed as local parameter to `exportTourbook()`

paetz (changes requested):

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookExporter.kt
chrgernoe: What is the alternative? I don't want to use Strings.

paetz: I don't mean the enum itself but only the semicolon

app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
```diff
@@ -46,8 +48,22 @@ class TourbookActivity : BaseNavigationActivity() {
 
     private lateinit var _db: DatabaseWrapper
     private lateinit var _availableYears: IntArray
+    private lateinit var _tourbookExporter: TourbookExporter
     private var _currentYear: Int = 0
     private var _tourbookType: TourbookType = TourbookType.eAscends
+    // Activity result handler
+    private val _export = registerForActivityResult(
```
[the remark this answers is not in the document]
chrgernoe: Not really. I named it now `_export|_importResultLauncher`. I got inspired by the Android Studio refactoring succestion.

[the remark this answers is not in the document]
paetz: Hm, I was just confused because you call `launch()` on that variable. And in the app we usually launch activities.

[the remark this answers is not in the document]
chrgernoe: This `registerForActivityResult` is the new way of getting back a result of an activity. The callback method, we are currently using all over the app, is deprecated. Therefore I switched at least this code to the new approach.

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookEntryVerbose.kt
chrgernoe: > Can `keys()` actually be private?

No, because it used in the `TourbookExporter` class.

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookExporter.kt
chrgernoe: Why that? It is a enum. Hence, it can only set to dedicated values?

chrgernoe: Original I wanted to pass this as local parameter but the problem is that the format is set in a different method than the actual export is done. Hence, I would need a variable to store this information anyway. So, I decided to store it directly in the object. The other approach would be to pass it through all methods, which is in my eyes very nasty. Or is there a smarter way to do this?

chrgernoe:
I realized most of your requests but there are still some unresolved review comments.

app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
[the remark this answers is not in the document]
paetz: I know. I was just questioning the naming.

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookExporter.kt
paetz: > Why that? It is a enum. Hence, it can only set to dedicated values?

It's not about the type but the fact that you can manipulate a state variable from outside. This may result in tricky bugs. You can just leave it as is, maybe I'll have a look at it again when doing the refactoring issue for next release.

app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
```diff
@@ -311,17 +326,35 @@ class TourbookActivity : BaseNavigationActivity() {
...
-        startActivityForResult(intent, IntentConstants.REQUEST_OPEN_TOURBOOK)
+        _importResultLauncher.launch(intent)
+    }
+
+    private fun _selectExportFormat() {
+        var mimeType = MIME_TYPE_JSON
+        DialogWidgetBuilder(this, R.string.export_format).apply {
+            setSingleChoiceItems(R.array.exportFormats, 0) {_, which ->
+                when (which) {
+                    0 -> _tourbookExporter.exportFormat = TourbookExporter.ExportFormat.eJSON
```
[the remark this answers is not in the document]
paetz: I guess, the cleanest way is to make TourbookExporter abstract and have dedicated `JSONTourbookExporter` and `CSVTourbookExporter`, respectively, as well as a dedicated importer class.
These concrete exporters would then handle their own export files types (mimetype).

However, this could be done as a follow-up if the scenario above works with the current approach.

paetz (changes requested):

app/src/main/java/com/yacgroup/yacguide/TourbookActivity.kt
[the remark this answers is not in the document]
chrgernoe: > Is this only executed when you actively select an item? Do we have a problem with the following:
> 
> 1. Open the dialog and select CSV (-> export format is set at exporter)
> 2. Cancel the dialog (negative button)
> 3. Open the dialog again and don't select anything (I guess, JSON (= idx 0). is default).
> 4. Confirm and choose your json file.
> 5. Now as we call `TourbookExporter.export()`, we want to export json but have we really set the export format or is it still CSV?
> 
> That's what I mean with "dangerous".

Good catch. Yes, it is as you described.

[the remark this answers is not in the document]
chrgernoe: > I guess, the cleanest way is to make TourbookExporter abstract and have dedicated `JSONTourbookExporter` and `CSVTourbookExporter`, respectively, as well as a dedicated importer class. These concrete exporters would then handle their own export files types (mimetype).
> 
> However, this could be done as a follow-up if the scenario above works with the current approach.

I fixed the issue in a more less nice way but agree whit your proposal.

paetz (approved):

PR #286, opened 2022-01-01
Global route search

### LINKED ISSUE #256: Global route search
Subtask of #251 

- Add menu item "Wege suchen" to views Country, Region, Sector, Rock, Route
- On click, a dialog should open with full text route search (similar to rock search)
- The found routes may be displayed on a SelectedRoutes view

Depending on #291

paetz:
Resolves #256

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/utils/SearchBarHandler.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/utils/SearchBarHandler.kt
@@ -33,6 +33,7 @@ class SearchBarHandler(searchBarLayout: ConstraintLayout,
                        checkBoxTitle: String,
                        checkBoxDefaultValue: Boolean,
                        private val _settings: SharedPreferences,
+                       private val _initCallback: (Boolean) -> Unit,
                        private val _updateCallback: (String, Boolean) -> Unit) {
 
     private var _namePart: String = ""
@@ -67,7 +68,7 @@ class SearchBarHandler(searchBarLayout: ConstraintLayout,
             })
         }
 
-        _updateCallback(_namePart, _onlyOfficialObjects)
+        _initCallback(_onlyOfficialObjects)
     }
 
     fun storeCustomSettings(key: String) {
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/yacgroup/yacguide/utils/SearchBarHandler.kt
```diff
@@ -33,6 +33,7 @@ class SearchBarHandler(searchBarLayout: ConstraintLayout,
                        checkBoxTitle: String,
                        checkBoxDefaultValue: Boolean,
                        private val _settings: SharedPreferences,
+                       private val _initCallback: (Boolean) -> Unit,
```
paetz, line 36: Note: This change was due to a little bug where `displayContent()` is called twice when starting `Rock-` or `RouteActivity`. However, it somehow belongs to this PR because searching `Alter Weg` on country level requires a lot of computation time (when building all widgets), and with the bug it would even take twice as much.

paetz: We should think of an issue for resolving this problem. Creating thousands of widgets (which is possible now with route search) is pretty time-consuming. Maybe there are more effective ways like creating only those widgets that are visible in the scroll view?

chrgernoe: Yeah. I already passed this topic when I implemented the `What's New` dialog. I think the answer is the `ListView` class which handles this for you.

chrgernoe (changes requested):
As you already mentioned, if the user search for a common name the UI basically hangs and on my device I get a message that the app does not response. On a device with less compute power it gets even worse. I think we need another solution for this.

paetz:
> As you already mentioned, if the user search for a common name the UI basically hangs and on my device I get a message that the app does not response. On a device with less compute power it gets even worse. I think we need another solution for this.

Ok, if this is a no-go, I suggest to decline this PR and move the issue to the next version. Switching to `ListView` would then be a necessary refactoring before creating this PR again. Do you agree?

PR #290, opened 2022-01-03
Updating country list on updating all via region manager

### LINKED ISSUE #285: Update country and region lists on updating all
In #281 we moved the update mechanism to a specialized region manager but forgot to add an opportunity to update country and region lists.

Todo: Update country list and all region lists of already downloaded countries in
A) the already existing menu point "Alle aktualisieren" or
B) in a separate menu point (because updating all sectors might be pretty time consuming in case you only want to get a newly added country.

paetz:
Resolves #285 

This PR adds update of country list to the update-all functionality in the region manager.

Note: The fact that updating is asynchronous leads to the (in my eyes) ugly callback implementation where we pass the next element's processing as a callback to the current one's update to be executed after that is finished.

In the future, we might think of refactoring this to Promise pattern.

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/RegionManagerActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/RegionManagerActivity.kt
@@ -17,47 +17,45 @@
 
 package com.yacgroup.yacguide
 
-import android.annotation.SuppressLint
 import android.os.Bundle
 import android.util.TypedValue
 import android.view.Menu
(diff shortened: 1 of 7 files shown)


chrgernoe (approved):
You are right, the code looks a bit nasty. According to the doc, the Kotlin way to handle async tasks are [coroutines](https://kotlinlang.org/docs/async-programming.html#coroutines). However, I'm not an expert for async programming techniques. I just had a brief look.

PR #292, opened 2022-01-03
Do not re-display rock and route lists on instantiating search bar

paetz:
Rock and route lists were displayed twice on creating the corresponding activity which doubles the time for time-consuming widgets creation.


chrgernoe (approved):
Was this the reason that the toast message was shown twice as I complained in #274?

paetz:
> Was this the reason that the toast message was shown twice as I complained in #274?

Indeed!

PR #307, opened 2022-01-27
Migrated country list to recycler view

### LINKED ISSUE #302: Migrate country list to RecyclerView
Subtask of #291

paetz:
Resolves #302

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/CountryActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/CountryActivity.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 Axel Paetzold
+ * Copyright (C) 2019, 2022 Axel Paetzold
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/app/src/main/java/com/yacgroup/yacguide/CountryActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/CountryActivity.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 Fabian Kantereit
+ * Copyright (C) 2019 Axel Paetzold
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -19,15 +19,15 @@ package com.yacgroup.yacguide
 
 import android.content.Intent
 import android.os.Bundle
-import android.view.View
-import android.widget.LinearLayout
+import androidx.recyclerview.widget.RecyclerView
 import com.yacgroup.yacguide.activity_properties.*
+import com.yacgroup.yacguide.list_adapters.CountryViewAdapter
 import com.yacgroup.yacguide.network.CountryParser
 import com.yacgroup.yacguide.utils.IntentConstants
-import com.yacgroup.yacguide.utils.WidgetUtils
 
 class CountryActivity : TableActivityWithOptionsMenu() {
 
+    private lateinit var _viewAdapter: CountryViewAdapter
     private lateinit var _updateHandler: UpdateHandler
 
     override fun onCreate(savedInstanceState: Bundle?) {
@@ -36,6 +36,9 @@ class CountryActivity : TableActivityWithOptionsMenu() {
         _updateHandler = UpdateHandler(this, CountryParser(db))
         properties = arrayListOf(RockSearchable(this), AscentFilterable(this))
 
+        _viewAdapter = CountryViewAdapter { countryName -> _onCountrySelected(countryName) }
+        findViewById<RecyclerView>(R.id.tableRecyclerView).adapter = _viewAdapter
+
         WhatsNewInfo(this).let {
             if (it.checkForVersionUpdate()) {
                 it.showDialog()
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/yacgroup/yacguide/UpdateHandler.kt
```diff
@@ -18,7 +18,9 @@
 package com.yacgroup.yacguide
 
 import android.app.Dialog
+import android.view.View
 import android.widget.ImageButton
+import android.widget.LinearLayout
```
paetz, line 23: Not needed

app/src/main/java/com/yacgroup/yacguide/database/Country.kt
```diff
@@ -18,9 +18,9 @@
...
 import androidx.room.Entity
+import androidx.room.PrimaryKey
 
-@Entity(primaryKeys = ["name"])
-class Country {
-
-    var name = ""
-}
+@Entity
+data class Country(
```
paetz, line 24: FYI: `data` classes allow property-wise object comparison by implementing an according `==` operator. I use this feature in the adapter to check the difference of list items (which are updated by RecyclerView if they differ). Example: Returning to the route activity while its ascent bitmask has changed (and it hence should be colored).

app/src/main/res/layout/content_table.xml
```diff
@@ -20,7 +20,7 @@
 <ScrollView
     xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="match_parent"
-    android:layout_height="match_parent"
+    android:layout_height="wrap_content"
```
paetz, line 23: ...otherwise there is no space left for the download button

app/src/main/res/layout/list_data_item.xml
```diff
@@ -0,0 +1,106 @@
...
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+  -->
+<LinearLayout
```
paetz, line 18: Note: This item is generic and will be used for all object types from country down to route. That's why there are so many text fields.

app/src/main/res/layout/activity_country.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  Copyright (C) 2019 Fabian Kantereit
```
paetz, line 3: Wrong name

app/src/main/java/com/yacgroup/yacguide/CountryActivity.kt
```diff
@@ -36,35 +36,29 @@ class CountryActivity : TableActivityWithOptionsMenu() {
...
-            layout.addView(_updateHandler.getDownloadButton { displayContent() })
-        }
+        _viewAdapter.submitList(countries)
+        _updateHandler.configureDownloadButton(countries.isEmpty()){ displayContent() }
+    }
+
+    private fun _onCountrySelected(countryName: String) {
+        val intent = Intent(this@CountryActivity, RegionActivity::class.java)
+        intent.putExtra(IntentConstants.CLIMBING_OBJECT_LEVEL, ClimbingObjectLevel.eRegion.value)
+        intent.putExtra(IntentConstants.CLIMBING_OBJECT_PARENT_NAME, countryName)
```
chrgernoe, line 61: Please use `apply`.

app/src/main/res/layout/button_download.xml
```diff
@@ -17,13 +17,16 @@
 
 <ImageButton
     xmlns:android="http://schemas.android.com/apk/res/android"
-    style="@style/AppTheme.Secondary.ImageButton"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
```
chrgernoe, line 20: Why did you change the button theme?

app/src/main/res/layout/list_view_content.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+  Copyright (C) 2019 Axel Paetzold
```
chrgernoe, line 4: Wrong year.

app/src/main/res/values/colors.xml
```diff
@@ -31,8 +31,8 @@
     <color name="colorOnPrimaryDark">@color/white</color>
 
     <!-- NOTE: Must be in sync with the colors in the WidgetUtils class. -->
-    <color name="colorSecondary">#9e9e9e</color>
-    <color name="colorSecondaryLight">#cfcfcf</color>
```
chrgernoe, line 35: Why did you change the color?

chrgernoe (changes requested):
The list items are now have a very high height. I looks a bit odd for the country names.

paetz:
> The list items are now have a very high height. I looks a bit odd for the country names.

You really think so? Whereas I find the old (=current) country appearance quite odd since it's much too small.

app/src/main/res/values/colors.xml
paetz: Because I added margin to the layout. Doing so, the old background color didn't really fit it.

app/src/main/res/layout/button_download.xml
paetz: That's what I already disliked about the old impl., that the download wasn't really visualized like a button. This way, I find it quite easier for the user.

chrgernoe:
> > The list items are now have a very high height. I looks a bit odd for the country names.
> 
> You really think so? Whereas I find the old (=current) country appearance quite odd since it's much too small for my fingers. Compromise: I'm gonna make the info text field on top invisible by default. So that item height is only 2 * currHeight instead of 3 * currHeight

Now, it looks OK. There is another small thing: Why is there a separator on top but no additional item?

app/src/main/res/layout/button_download.xml
chrgernoe: Please do not change the style individual for each item. Use styles instead. Otherwise it gets a mess again. I tried to clean this up with PR #217.

app/src/main/res/values/colors.xml
chrgernoe: OK

chrgernoe:
In general, please have a look here: https://www.gnu.org/licenses/gpl-howto.en.html

chrgernoe (changes requested):

paetz:
> > > The list items are now have a very high height. I looks a bit odd for the country names.
> > 
> > 
> > You really think so? Whereas I find the old (=current) country appearance quite odd since it's much too small for my fingers. Compromise: I'm gonna make the info text field on top invisible by default. So that item height is only 2 * currHeight instead of 3 * currHeight
> 
> Now, it looks OK. There is another small thing: Why is there a separator on top but no additional item?

This is by intent to visualize the table start since there may be other text fields above like rock status.

chrgernoe (approved):

PR #311, opened 2022-02-01
Migrated region list to recycler view

### LINKED ISSUE #303: Migrate region list to RecyclerView
Subtask of #291

paetz:
Resolves #303

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/RegionActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/RegionActivity.kt
@@ -17,19 +17,17 @@
 
 package com.yacgroup.yacguide
 
-import android.annotation.SuppressLint
 import android.content.Intent
-import android.graphics.Color
 import android.os.Bundle
-import android.view.View
-import android.widget.LinearLayout
+import androidx.recyclerview.widget.RecyclerView
 import com.yacgroup.yacguide.activity_properties.*
+import com.yacgroup.yacguide.list_adapters.RegionViewAdapter
 import com.yacgroup.yacguide.network.RegionParser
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/yacgroup/yacguide/RegionActivity.kt
```diff
@@ -17,57 +17,43 @@
...
-        for (region in regions) {
-            val onClickListener = View.OnClickListener {
-                val intent = Intent(this@RegionActivity, SectorActivity::class.java)
-                intent.putExtra(IntentConstants.CLIMBING_OBJECT_LEVEL, ClimbingObjectLevel.eSector.value)
-                intent.putExtra(IntentConstants.CLIMBING_OBJECT_PARENT_ID, region.id)
-                intent.putExtra(IntentConstants.CLIMBING_OBJECT_PARENT_NAME, region.name)
-                startActivity(intent)
-            }
-            var icon = getString(R.string.empty_box)
-            var bgColor = Color.WHITE
```
chrgernoe, line 59: Instead hard-coding a color here, it would be good if a central color is defined. So, it makes it easier to re-style the app.

chrgernoe (approved):

app/src/main/java/com/yacgroup/yacguide/RegionActivity.kt
paetz: > Instead hard-coding a color here, it would be good if a central color is defined. So, it makes it easier to re-style the app.

This has been removed :)

chrgernoe: Oh, right :-)

Issue #248, opened 2021-07-16
As a user, I want to be able to set a region or country as default


ckummer:
When opening the app, the standard region or country should pop up. E. g., when the region is set to "Sächsische Schweiz", it should show the subregions of the selected region, that is, Brand, Steine, ...

paetz:
This implies to have some kind of "parent button" which is not available yet (since we haven't solved the problem of cyclic view calls). The "back button" returns to the view that has been visited just before. 
Adding this feature without a "parent button" would mean that you will be never able to get to the parent country of your standard region.

ckummer:
I agree, this implies to have a home/parent button to or something the like. An alternative approach would be add a region manager similar to OSMand where you add regions/countries and switch regions using a burger menu or something the like. However, this feels also like an usability enhancement, but selecting the country and the region over and over again is annoying.

paetz:
https://stackoverflow.com/questions/15697735/how-to-change-start-activity-dynamically

paetz:
Depending on #289

paetz:
The default region selection should be available in the region manager.
Suggestion:
- on clicking a region name:
-- restore the current default region (incl. de-highlighting)
-- store according region id in shared preferences
-- show Toast like "<Name> als Standardgebiet festgelegt"
-- highlight according item
- on clicking a highlighted region:
-- remove default region id in shared preferences
-- show a Toast like "<Name> als Standardgebiet entfernt"
-- de-highlight according item

PR #312, opened 2022-02-01
Added start-up activity

### LINKED ISSUE #289: Add start-up activity
Add a start-up activity with blank screen containing only the Yac Logo.
There should be a time-out (maybe 1 sec) before switching to `CountryActivity`.

(In the future, we may add automatic country list update within that activity.)

paetz:
Resolves #289


chrgernoe (approved):
Without doing anything, it is a bit useless but OK.

paetz:
> Without doing anything, it is a bit useless but OK.

It is a requirement for #248

PR #318, opened 2022-02-07
Added unit tests for country dao

### LINKED ISSUE #317: CountryDao unit tests

paetz:
Resolves #317


paetz:
Do you know what to apply to the CI to make it work?

chrgernoe:
> Do you know what to apply to the CI to make it work?

I guess, locally it it works. I will have a look.

chrgernoe:
You broke the [ExampleUnitTest](app/src/test/java/com/yacgroup/yacguide/ExampleUnitTest.kt). I think because removed the test implementation statements  in the `build.gradle` file.

chrgernoe:
Why did you implement a instrumentation test instead of a simple unit test? The documentation says:

> We recommend using instrumented tests only in cases where you must test against the behavior of a real device.

See https://developer.android.com/training/testing/instrumented-tests.

paetz:
> > Do you know what to apply to the CI to make it work?
> 
> I guess, locally it it works. I will have a look.

Yes, it works locally.

paetz:
> You broke the [ExampleUnitTest](app/src/test/java/com/yacgroup/yacguide/ExampleUnitTest.kt). I think because removed the test implementation statements in the `build.gradle` file.

Ah, I see. I'm gonna remove this file.

paetz:
> Why did you implement a instrumentation test instead of a simple unit test? The documentation says:
> 
> > We recommend using instrumented tests only in cases where you must test against the behavior of a real device.
> 
> See https://developer.android.com/training/testing/instrumented-tests.

Because I test the Room database. This requires an Instrumented test since it can only run on a device.

chrgernoe:
I found a solution and will push the GitHub action the next days. Fortunately, somebody solved the issue already because for newer API versions it is not so easy or just impossible to run the emulator inside a Docker container when using GitHub Action because these emulator images need hardware acceleration, which is not enabled. However, when the Action is run on macOS (which is available on GitHub) the hardware acceleration is enabled and so the instrumentation test can be run.

paetz:
> I found a solution and will push the GitHub action the next days. Fortunately, somebody solved the issue already because for newer API versions it is not so easy or just impossible to run the emulator inside a Docker container when using GitHub Action because these emulator images need hardware acceleration, which is not enabled. However, when the Action is run on macOS (which is available on GitHub) the hardware acceleration is enabled and so the instrumentation test can be run.

You are the man!

chrgernoe (approved):
I installed the workflow and tested it on this branch. Therefore I had to rebase this branch.

PR #320, opened 2022-02-08
Migrated route list to recycler view

### LINKED ISSUE #306: Migrate route list to RecyclerView
Subtask of #291

paetz:
Resolves #306


chrgernoe (changes requested):
The last entry is truncated on my screen. Especially this can be seen if there is an alternative name shown as in the Czech Republic. Basically this issues occurs on every level (country, region, sector, rock) because the last entry has not the full length.

paetz:
> The last entry is truncated on my screen. Especially this can be seen if there is an alternative name shown as in the Czech Republic. Basically this issues occurs on every level (country, region, sector, rock) because the last entry has not the full length.

I've fixed it but needed to remove the list margin for this. But since you had criticized this anyways, it should be fine for you :)

chrgernoe (approved):

PR #321, opened 2022-02-10
Possibility to set default region

### LINKED ISSUE #248: As a user, I want to be able to set a region or country as default

paetz:
Resolves #248

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/RegionManagerActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/RegionManagerActivity.kt
@@ -72,7 +72,9 @@ class RegionManagerActivity : BaseNavigationActivity() {
         val layout = findViewById<LinearLayout>(R.id.tableLayout)
         layout.removeAllViews()
 
-        val defaultRegionId = _customSettings.getInt(getString(R.string.default_region), R.integer.unknown_id)
+        val defaultColor = ContextCompat.getColor(this, R.color.colorSecondaryLight)
+        val highlightedColor = ContextCompat.getColor(this, R.color.colorAccentLight)
+        val defaultRegionId = _customSettings.getInt(getString(R.string.default_region_key), R.integer.unknown_id)
         val defaultRegionIcon = getString(R.string.startup_arrow)
         var currentCountryName = ""
         val regions = _db.getNonEmptyRegions() // already sorted by country name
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/yacgroup/yacguide/RegionManagerActivity.kt
```diff
@@ -159,4 +177,21 @@ class RegionManagerActivity : BaseNavigationActivity() {
             _updateHandler.finish()
         }
     }
+
+    private fun _selectDefaultRegion(regionId: Int) {
+        val key = getString(R.string.default_region)
```
chrgernoe, line 182: Why is the German word `Standardgebiet` used internal as settings key? Because this string is defined in a file which is used to translate the app into other languages, this key changes for each language. It makes no sense and even can lead to the fact that the settings get lost, if the language is changed. I know, at the moment this cannot happen because the app is only available in German but in general it should be avoided. Why not just defining it in the `setup.xml`?

chrgernoe (changes requested):
This feature shows me that we have to explain carefully inside the next release notes what has been changed otherwise the user will only accidentally find this feature.

paetz:
> This feature shows me that we have to explain carefully inside the next release notes what has been changed otherwise the user will only accidentally find this feature.

Just like the pinning/archiving feature of Threema which I had not known before you told me.

app/src/main/java/com/yacgroup/yacguide/RegionManagerActivity.kt
```diff
@@ -78,8 +86,18 @@ class RegionManagerActivity : BaseNavigationActivity() {
                 currentCountryName = region.country.orEmpty()
             }
 
+            var regionDispName = region.name.orEmpty()
+            var bgColor = Color.WHITE
```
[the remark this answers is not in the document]
paetz: Yes, of course. This should be `R.color.colorSecondaryLight`. Strange that I missed that.

app/src/main/res/values/strings.xml
```diff
@@ -160,6 +163,8 @@
     <string name="only_official_summits">Nur offizielle Gipfel</string>
     <string name="only_official_routes">Nur offizielle Wege</string>
 
+    <string name="default_region">Standardgebiet</string>
```
chrgernoe, line 166: I think this definition is not used anymore after your last change.

chrgernoe (approved):

PR #323, opened 2022-02-17
Enabled global route search

### LINKED ISSUE #256: Global route search

paetz:
Resolves #256


chrgernoe (approved):
Nice! Even for a large number of results, it is very fast.

PR #324, opened 2022-02-18
Drop support for API version <26

### LINKED ISSUE #322: Drop support for API version <26 (Android 8.0)
According to the Google Play Store statistics, we have only one device with Android 5.1 (API 22) and two devices on Android 8.1 (API 27). I guess this one device with Android 5.1 is not even in use anymore. Therefore, I suggest to drop the support for the API <26 (Android 8.0) because we already have some code the is limited by this lower API levels. 

@paetz: What do you think?
@JoergB2: As far as I know, the support for API 22 was kept only for you. Would it be OK for you to drop this support or do you still use this device?

chrgernoe:
Fixes #322

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -35,7 +35,7 @@ android {
     compileSdkVersion 31
     defaultConfig {
         applicationId "com.yacgroup.yacguide"
-        minSdkVersion 22
+        minSdkVersion 27
         targetSdkVersion 31
         testInstrumentationRunner 'androidx.test.runner.AndroidJUnitRunner'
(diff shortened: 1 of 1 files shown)


app/build.gradle
```diff
@@ -35,7 +35,7 @@ android {
     compileSdkVersion 31
     defaultConfig {
         applicationId "com.yacgroup.yacguide"
-        minSdkVersion 22
+        minSdkVersion 27
```
paetz, line 38: The branch name states to drop support below api level 26. So why min. 27 here?

paetz (approved):

Issue #262, opened 2021-11-23
Remove warnings and refactor


paetz:
- Remove warnings, e.g. deprecated APIs
- Rename all `Ascend...` stuff to `Ascent...`
- Migrate to view bindings since kotlin-android-extensions is deprecated
- Remove deprecated code
- `inline` where possible
- Summarize SQL macros that are often used, e.g. `VIA_ROUTES_REGION = "$VIA_ROUTES_ROCK $VIA_ROCKS_SECTOR $VIA_SECTORS_REGION"`
- Remove all `!!` occurrences by proper error handling
- Use separate string keys defined in `setup.xml` for SharedPreferences keys

chrgernoe:
I already noticed that but first wanted to finish #213 because there will probably even more after the API change.

PR #331, opened 2022-02-22
Display ascent list below route description

### LINKED ISSUE #250: As a user, I want to see comments to routes as well as my ascents below the route description
As a user, I want to see comments from other users as well as my ascents below the description. These should be separated into two sections and should be either collapsable or in tabs.

paetz:
Resolves #250

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookExporter.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookExporter.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 Axel Pätzold
+ * Copyright (C) 2019, 2022 Axel Pätzold
  * Copyright (C) 2021 Christian Sommer
  *
  * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookExporter.kt
```diff
@@ -146,28 +146,17 @@ class TourbookExporter(
         val jsonAscends = JSONArray(jsonString)
         _db.deleteAscends()
         _db.deletePartners()
-        val ascends = mutableListOf<Ascend>()
+        val ascents = mutableListOf<Ascend>()
```
chrgernoe, line 149: I think the class `Ascend` also needs to be renamed to `Ascent`. But maybe in a separate PR.

chrgernoe (approved):
Maybe the partner list could also be shown in the ascent overview of the route.

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookExporter.kt
paetz: I know. This is part of a refactoring issue (#262). However, I believe, renaming this class is not possible without database structure migration (because it is the name of a database entity).

paetz:
> Maybe the partner list could also be shown in the ascent overview of the route.

I think, this might become ugly once you have more than one partner and for small devices. Furthermore, the detailed ascent view would only have the notes as additional info in that case which makes it almost (but not completely) obsolete.

PR #334, opened 2022-02-24
Added protection filter for routes

### LINKED ISSUE #272: Add filter Protection to route filter dialog
Subtask of #251

paetz:
Resolves #272


chrgernoe (approved):
What, if the user want to search only for poor protected routes. Would it make sense to implement a from-to filter in a separate PR?

PR #347, opened 2022-03-14
Added subjective grade filter for routes

### LINKED ISSUE #271: Add filter SubjectiveGrade to route filter dialog
Subtask of #251 

Important note: This is NOT a filter for the official grade but the subjective grades from user comments.
It should be possible to set a grade range in the dialog.

paetz:
Resolves #271

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -80,6 +80,8 @@
     <string name="route_search">Weg suchen</string>
     <string name="partner_search">Partner suchen</string>
     <string name="rock_relevance_filter">Bedeutung: mindestens</string>
+    <string name="route_grade_filter">Schwierigkeitseinschätzung:</string>
+    <string name="route_grade_filter_to">bis</string>
     <string name="route_quality_filter">Qualität: mindestens</string>
     <string name="route_protection_filter">Absicherung: mindestens</string>
     <string name="route_drying_filter">Abtrocknung: mindestens</string>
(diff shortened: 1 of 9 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -80,6 +80,8 @@
     <string name="route_search">Weg suchen</string>
     <string name="partner_search">Partner suchen</string>
     <string name="rock_relevance_filter">Bedeutung: mindestens</string>
+    <string name="route_grade_filter">Schwierigkeitseinschätzung:</string>
```
chrgernoe, line 83: We should name it "Mittlere Schwierigkeitseinschätzung" because you calculate the average.

chrgernoe (changes requested):
I would questioning, if the average is really the best function to use because the subjective grade  depends much on the climbing skills which can lead to outliers and can have a big influence on the average. Maybe the median or mode is more suitable. See https://en.wikipedia.org/wiki/Mode_(statistics).

app/src/main/res/values/strings.xml
paetz: In this case, we would need this for each filter

paetz:
> I would questioning, if the average is really the best function to use because the subjective grade depends much on the climbing skills which can lead to outliers and can have a big influence on the average. Maybe the median or mode is more suitable. See https://en.wikipedia.org/wiki/Mode_(statistics).

I question you questioning :)
- Mode may be ambiguous.
- Both (Mode and Median) are not available in SQL.
- The only problem with average are outliers. But according to our experience, outliers are extremely rare on sandsteinklettern.de and could be removed by either considering only well-rated comments or manually remove trash comments on sandsteinklettern.de

app/src/main/res/values/strings.xml
chrgernoe: Maybe.

chrgernoe (approved):

PR #359, opened 2022-11-28
Unified ascend-ascent naming clash

paetz:
Changed all occurrences of `ascent` to `ascend`.

I know `ascend` is wrong and it should be `ascent`. But since we cannot change the db table names without migration logic, I changed all occurrences to `ascend`. It should be equal throughout the code base.

Once we need a db migration anyway someday, we may change everything to `ascent`.


chrgernoe:
I would not do this. You change a lot of code only because you cannot change this one database class. In my eyes it is not worth doing this. I would change/leave everything into the `ascent` except for the code where you cannot do the change and add just a comment in the code. In addition you can create an issue where we collect all things which need to be done, if a database change is necesary.

paetz:
Ok, I'm gonna wait until we migrate the db anyways to apply this name change, too.

PR #362, opened 2022-11-30
Added modern float action buttons

### LINKED ISSUE #298: Replace "+" button in partner and description activities by FloatingActionButton
Depends on #291 , #366

See https://github.com/android/views-widgets-samples/blob/main/RecyclerViewKotlin/app/src/main/res/layout/activity_main.xml

paetz:
Resolves #298


app/src/main/res/layout/route_description.xml
```diff
@@ -91,13 +91,33 @@
...
+
+        <com.google.android.material.floatingactionbutton.FloatingActionButton
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_margin="16dp"
+            android:onClick="enterAscend"
+            android:src="@android:drawable/ic_input_add"
+            app:layout_constraintRight_toRightOf="parent"
+            app:layout_constraintTop_toTopOf="parent"
+            app:layout_constraintBottom_toBottomOf="parent"/>
```
chrgernoe, line 118:
```suggestion
            app:layout_constraintBottom_toBottomOf="parent"
            style="@style/floatingActionButtonStyle"/>
```
Add a style to file `style.xml`:
```xml
<style name="floatingActionButtonStyle">
    <item name="tint">@color/colorOnPrimary</item>
    <item name="backgroundTint">@color/colorPrimary</item>
</style>
```
I'm not fully happy with the style but for now it is OK.

chrgernoe (changes requested):
All apps I know place this kind of button in the right bottom corner. I would also place it there. At the current position it somehow looks not OK.

paetz:
> All apps I know place this kind of button in the right bottom corner. I would also place it there. At the current position it somehow looks not OK.

This is not possible right now since it would then cover the Edit and Delete buttons in the Partner list. Solution would be to make Edit/Delete available via Swipe and remove those buttons

chrgernoe:
Adding via gestures is in my eyes not very intuitive. Is it possible to hide the button by default and make it visible only in this view?

paetz:
> Adding via gestures is in my eyes not very intuitive. Is it possible to hide the button by default and make it visible only in this view?

You misunderstood me. Solution is Edit/Delete on swiping while the ActionButton for adding is in the right down corner (overlaying the partner list in that area), just as you desired.

chrgernoe:
Deleting via swipe gesture is fine for me but editing via swipe is also not so nice in my eyes. Another suggestion would be the long press on a partner. This would then open a menu for edit and delete. See Threema app on the chat list.

Just that I understand it right. The action button is used in the partner view also to add a new partner? 

> overlaying the partner list in that area

I can live with that. It is the same in other apps.

chrgernoe (approved):

PR #364, opened 2022-12-09
Upgrade Kotlin version to 1.7.20

chrgernoe:
Fixed a deprecation warning that occurs with this upgrade. See https://developer.android.com/guide/navigation/navigation-custom-back

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/BaseNavigationActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/BaseNavigationActivity.kt
@@ -21,6 +21,7 @@ import android.content.Intent
 import android.os.Bundle
 import android.view.Menu
 import android.view.MenuItem
+import androidx.activity.OnBackPressedCallback
 import androidx.appcompat.app.AppCompatActivity
 import androidx.appcompat.app.ActionBarDrawerToggle
 import androidx.appcompat.widget.Toolbar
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/yacgroup/yacguide/BaseNavigationActivity.kt
```diff
@@ -29,12 +30,15 @@ import androidx.drawerlayout.widget.DrawerLayout
...
 
 abstract class BaseNavigationActivity : AppCompatActivity(), NavigationView.OnNavigationItemSelectedListener {
-    override fun onBackPressed() {
-        val drawerLayout: DrawerLayout = findViewById(R.id.drawer_layout)
-        if (drawerLayout.isDrawerOpen(GravityCompat.START)) {
-            drawerLayout.closeDrawer(GravityCompat.START)
-        } else {
-            super.onBackPressed()
+
+    private val onBackPressedCallback: OnBackPressedCallback = object: OnBackPressedCallback(true) {
```
paetz, line 34: This does not need to (and should not) be a class member. Please define it within `onCreate()` where you use it:
```
onCreate() {
    ...
    onBackPressedDispatcher.addCallback(this, OnBackPressedCallback(true) {
        handleOnBackPressed() {
            ....
        }
    })
}

**Review by paetz — CHANGES_REQUESTED** [review]

File: app/src/main/java/com/yacgroup/yacguide/BaseNavigationActivity.kt [hunk 2]
```diff
@@ -62,7 +51,15 @@ abstract class BaseNavigationActivity : AppCompatActivity(), NavigationView.OnNa
 
         navView.setNavigationItemSelectedListener(this)
 
-        onBackPressedDispatcher.addCallback(this, onBackPressedCallback)
+        onBackPressedDispatcher.addCallback(this, object: OnBackPressedCallback(true) {
+            override fun handleOnBackPressed() {
+                if (drawerLayout.isDrawerOpen(GravityCompat.START)) {
```
chrgernoe, line 56: @paetz: Here I'm bit unsure if this is a good programming style.  `drawerLayout` is captured on a closure.

paetz: I have no clue, either.

paetz (approved):

app/src/main/java/com/yacgroup/yacguide/BaseNavigationActivity.kt
chrgernoe: I'm also a bit surprised that this works anyway because you declare a callback function which is unrelated to the `onCreate` context. However, if you would use the `val drawerLayout = ...` you get a warning that the variable is shadowed by the outer scope. Nevertheless, it works :-).

PR #369, opened 2022-12-14
Update country and region list automatically on app launch

### LINKED ISSUE #310: Automatic update of countries and regions
Update country list and all region lists automatically on start-up (as part of the start-up activity, see #289).
If updating takes too long (--> timeout), just switch over to the database without updating and show a Toast that it did not work.

Depends on #289

paetz:
Resolves #310 

This again removes some complexity and code by updating countries and regions at once (which is not expensive regarding the little amount of data).


chrgernoe (approved):

chrgernoe:
Can you please check, if issue #370 is relevant here.

PR #376, opened 2022-12-28
Make sector updates cancelable

### LINKED ISSUE #353: Make sector updates cancelable
Especially when hitting "Update all" by accident, a possibility to cancel this would be nice.

paetz:
Resolves #353 

Note: I've changed the update error handling so that a region update is immediately cancelled if an error occurs. However, for Update-All-Requests, I preserved the handling of continue updating the other regions and omit only the failed one.

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/UpdateHandler.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/UpdateHandler.kt
@@ -19,6 +19,7 @@ package com.yacgroup.yacguide
 
 import android.app.Dialog
 import android.view.View
+import android.widget.Button
 import android.widget.ImageButton
 import android.widget.TextView
 import android.widget.Toast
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/yacgroup/yacguide/UpdateHandler.kt
```diff
@@ -55,11 +57,17 @@ class UpdateHandler(private val _activity: AppCompatActivity,
     }
 
     override fun onUpdateFinished(exitCode: ExitCode) {
-        _exitCodes.add(exitCode)
-        _onUpdateFinished()
-
-        if (!_isRecurringUpdate) {
+        if (exitCode == ExitCode.ABORT) {
```
paetz, line 60: I know this if branching is super-ugly but I haven't found a nicer way.

chrgernoe (approved):

PR #397, opened 2023-01-19
Encircle launch icon after finishing automatic update

### LINKED ISSUE #375: Add visual indicator for finished update on app launch
Once the automatic silent update on app launch has finished, the yac icon should change somehow to show this, e.g. by encircling the logo.

paetz:
Resolves #375

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022 Axel Paetzold
+ * Copyright (C) 2022, 2023 Axel Paetzold
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -22,6 +22,7 @@ import android.content.Intent
 import android.content.SharedPreferences
 import android.os.Bundle
 import android.os.CountDownTimer
+import android.widget.ImageView
 import androidx.appcompat.app.AppCompatActivity
 import com.yacgroup.yacguide.database.DatabaseWrapper
 import com.yacgroup.yacguide.network.CountryAndRegionParser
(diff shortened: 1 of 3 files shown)


chrgernoe:
Cool would be, if the circle is drawn clock wise while loading instad of just appearing.

chrgernoe (changes requested):
You need to provide two versions, one for the main and the other for the dev release because you just override the the "Dev" below the icon. I just realized that this is already the case in the current release.

paetz:
> Cool would be, if the circle is drawn clock wise while loading instead of just appearing.

yes, of course but this would be pretty difficult to implement.

paetz:
> You need to provide two versions, one for the main and the other for the dev release because you just override the the "Dev" below the icon. I just realized that this is already the case in the current release.

? I don't understand. Dev- or stable version shouldn't matter. The launch screen icon is the same for both versions.

chrgernoe:
> > Cool would be, if the circle is drawn clock wise while loading instead of just appearing.
> 
> yes indeed, but this would be pretty difficult to implement.

It is a nice gimmick. Since I have seen apps which have something similar, I guess there you probably find something in the internet.

chrgernoe:
> > You need to provide two versions, one for the main and the other for the dev release because you just override the the "Dev" below the icon. I just realized that this is already the case in the current release.
> 
> ? I don't understand. Dev- or stable version shouldn't matter. The launch screen icon is the same for both versions.

Actual not because we also have different icons for both versions in the launcher. And as I wrote, in the dev version the icon has the word dev in it until you override it.

paetz:
> > > Cool would be, if the circle is drawn clock wise while loading instead of just appearing.
> > 
> > 
> > yes indeed, but this would be pretty difficult to implement.
> 
> It is a nice gimmick. Since I have seen apps which have something similar, I guess there you probably find something in the internet.

If you'd like to have it, you may create a ticket. But keep in mind that
a) the update process at launch is pretty fast, so it might look odd
b) we do not send any status updates when updating countries & regions currently. If changing this, it would have an impact on the "update-all" feature in the region manager.

paetz:
> > > You need to provide two versions, one for the main and the other for the dev release because you just override the the "Dev" below the icon. I just realized that this is already the case in the current release.
> > 
> > 
> > ? I don't understand. Dev- or stable version shouldn't matter. The launch screen icon is the same for both versions.
> 
> Actual not because we also have different icons for both versions in the launcher. And as I wrote, in the dev version the icon has the word dev in it until you override it.

Then we have different dev versions :) On my device (dev version 0.6) the icon on launching is the normal one without the "dev" lettering underneath.

chrgernoe (approved):

PR #398, opened 2023-02-04
Fixed incorrect handling with default region id

paetz:
This resolves the issue that no countries are shown in start-up which was due to a resources bug.

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
@@ -59,8 +59,9 @@ class LaunchActivity : AppCompatActivity() {
     }
 
     private fun _switchToDatabase() {
-        val defaultRegionId = _customSettings.getInt(getString(R.string.default_region_key), R.integer.unknown_id)
-        val intent = if (defaultRegionId == R.integer.unknown_id)
+        val invalidId = resources.getInteger(R.integer.default_region_id)
+        val defaultRegionId = _customSettings.getInt(getString(R.string.default_region_key), invalidId)
+        val intent = if (defaultRegionId == invalidId)
             Intent(this, CountryActivity::class.java)
         else
             Intent(this, SectorActivity::class.java).apply {
--- a/app/src/main/java/com/yacgroup/yacguide/PreferencesActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/PreferencesActivity.kt
@@ -103,28 +103,40 @@ class PreferencesActivity : BaseNavigationActivity() {
     }
 
     private fun _resetCustomSettings() {
-        val editor = _customSettings.edit()
-        for ((_, keyPair) in _settingKeysMap) {
-            editor.putBoolean(getString(keyPair.first), resources.getBoolean(keyPair.second))
-        }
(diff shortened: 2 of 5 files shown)


app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
```diff
@@ -59,8 +59,9 @@ class LaunchActivity : AppCompatActivity() {
     }
 
     private fun _switchToDatabase() {
-        val defaultRegionId = _customSettings.getInt(getString(R.string.default_region_key), R.integer.unknown_id)
```
paetz, line 62: This was the bug: Accessing `R.integer.unknown_id` which does not represent the value itself but rather the id of the according xml resource.

app/src/main/java/com/yacgroup/yacguide/PreferencesActivity.kt
```diff
@@ -103,28 +103,42 @@ class PreferencesActivity : BaseNavigationActivity() {
...
+            for ((_, keyPair) in _settingKeysMap) {
+                putBoolean(getString(keyPair.first), resources.getBoolean(keyPair.second))
+            }
+            putInt(
+                getString(R.string.lead),
+                ContextCompat.getColor(context, R.color.color_lead))
+            putInt(
+                getString(R.string.follow),
+                ContextCompat.getColor(context, R.color.color_follow))
+            putInt(
```
paetz, line 117: Another bugfix: Resetting the default region id on resetting the app.

app/src/main/java/com/yacgroup/yacguide/database/DatabaseWrapper.kt
```diff
@@ -29,7 +29,7 @@ class DatabaseWrapper(context: Context) {
 
     companion object {
         const val UNKNOWN_NAME = "???"
-        const val INVALID_ID = R.integer.unknown_id
+        const val INVALID_ID = -1
```
paetz, line 32: Same bug here: You'd need `resources.getInteger(R.integer.unknown_id)` to retrieve the value. However, the database should not hold information about the calling context, so I switched back to an id separate from the default region id.

app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
chrgernoe: Why was this not a problem for higher AP versions?

chrgernoe (approved):

app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
paetz: The problem is that accessing raw resource `R.integer.unknown_id` twice seems to give different results for lower API levels.

paetz: I don't think it's API level related. I guess, I (or my emulator, respectively) had an older Yacguide version installed before, in which the raw resource id R.integer.unknown_id was different from a new app version. But since the old one was stored and never reset to -1, the if branch compared two different resource ids which led to the behavior that the app wants to load a region with an invalid id as start-up region.

PR #401, opened 2023-02-08
Migrate tourbook to nested recycler view

### LINKED ISSUE #382: Investigate sensibility of nested RecyclerView for Tourbook and RegionManager
A nested RecyclerView would be another (maybe cleaner?) solution for sectioned lists in Tourbook- and RegionManagerActivity. 
Check this out regarding code cleanness and performance.

paetz:
Sorry for this huge PR but using nested recycler view seems to be the only way to fix tourbook's inconsistency. And it's the much cleaner approach anyways.
Migration of `RegionManager`'s list will be done in a follow-up PR.

Resolves #382

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/list_adapters/TourbookAscendViewAdapter.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/list_adapters/TourbookAscendViewAdapter.kt
@@ -1,0 +1,95 @@
+/*
+ * Copyright (C) 2023 Axel Paetzold
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/yacgroup/yacguide/list_adapters/TourbookAscendViewAdapter.kt
```diff
@@ -0,0 +1,95 @@
+/*
```
paetz, line 1: I'm currently working on unifying all list adapters into one to reduce code duplication and split the code better into the responsible classes. This means, that this class will be removed again later.

chrgernoe (approved):

PR #406, opened 2023-02-12
Get rid of database schema export warning

chrgernoe:


app/schemas/com.yacgroup.yacguide.database.AppDatabase/1.json
```diff
@@ -0,0 +1,516 @@
...
+          "autoGenerate": false
+        },
+        "indices": [],
+        "foreignKeys": []
+      }
+    ],
+    "views": [],
+    "setupQueries": [
+      "CREATE TABLE IF NOT EXISTS room_master_table (id INTEGER PRIMARY KEY,identity_hash TEXT)",
+      "INSERT OR REPLACE INTO room_master_table (id,identity_hash) VALUES(42, '8303ae24eca627fcaebc04dd60fd4fd1')"
```
paetz, line 513: Why do you insert value 42 for `id` and where does this hash come from?

app/src/main/java/com/yacgroup/yacguide/database/AppDatabase.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 Fabian Kantereit
+ * Copyright (C) 2019 Axel Paetzold
```
paetz, line 2: Please add yourself for year 2023 since you've added `exportSchema = true`

app/schemas/com.yacgroup.yacguide.database.AppDatabase/1.json
```diff
@@ -0,0 +1,516 @@
...
+            "name"
+          ],
+          "autoGenerate": false
+        },
+        "indices": [],
+        "foreignKeys": []
+      },
+      {
+        "tableName": "Region",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` INTEGER NOT NULL, `name` TEXT, `country` TEXT, PRIMARY KEY(`id`))",
```
paetz, line 29: All the field types should correspond to the types in the according database classes, e.g. `'name' TEXT` should correspond to `name: String?` in the `Region` class definition. 
What if you change this definition here without changing the class definition? 
- In my eyes the definition should be only in one single place. Having a class definition and a schema definition of one and the same database table is quite odd.
- I think, it's good to have a versioned database but I don't see the need of this schema since the sql statements are created automatically from the class definitions.

paetz (changes requested):

paetz:
The "changes requested" is not really a change request but rather a clarification request.

app/src/main/java/com/yacgroup/yacguide/database/AppDatabase.kt
chrgernoe: I just fixed the original author. I did just a small change. I can add me on addition to you.

app/schemas/com.yacgroup.yacguide.database.AppDatabase/1.json
chrgernoe: - The file is exported automatically during build. Hence, any user-change will be overridden.  
- In the doc they stated as good practice to have this file but we can also set the `exportSchema = false`.
- We cannot influence the content of this file. I think they have a reason why they do it like this.

paetz: Ah, I see. Then please

- [ ] Add the schemas folder to the `.gitignore` file

app/src/main/java/com/yacgroup/yacguide/database/AppDatabase.kt
paetz: I would add you since you have changed the app's behavior on building. If you only rename stuff, I'd agree to not change the Copyright header.

paetz (approved):

PR #407, opened 2023-02-13
Unified list adapters to generic list adapter

paetz:
This is the 1st step of list adapter unification. 
- replaced almost all concrete adapters by generic `ListAdapter<T>`
- did some timing measurements: The new solution is only insignificantly slower than the old one

The 2nd (follow-up) step will contain:
- pass `SwipeController` to `SectionViewAdapter`
- migrate list adapter of region manager to `SectonViewAdapter`, containing subadapter of type `ListAdapter<Region>`
- Remove `BaseViewAdapter` and `BaseViewItem`


app/src/main/java/com/yacgroup/yacguide/utils/VisualUtils.kt
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+package com.yacgroup.yacguide.utils
+
+import android.content.Context
+import android.content.SharedPreferences
+import androidx.core.content.ContextCompat
+import com.yacgroup.yacguide.R
+
+class VisualUtils(
```
paetz, line 25: The reason for having this class is:
When creating the RecyclerView list items, you'd need to call ContextCompat methods (like `ContextCompat.getColor()` e.g.) for every item. Using `VisualUtils`, the according values are set on activity creation and no method calls are needed when iterating over the item list.

chrgernoe (approved):

PR #408, opened 2023-02-15
Use generic list adapter for region manager list

paetz:
Last PR to complete the migration to nested recycler views.


paetz:
> You somehow break the region manager selection. The arrow that marks the default region can now be on more than one region. The function behind seems to work correctly but visually it is not as expected.

Thanks for catching this. I've only tested with one country but it of course does not work for selecting between multiple countries.

chrgernoe (approved):

PR #409, opened 2023-02-15
Make deprecated package manager API compatible

chrgernoe:


app/src/main/java/com/yacgroup/yacguide/extensions/PackageManager.kt
```diff
@@ -0,0 +1,40 @@
...
+
+package com.yacgroup.yacguide.extensions
+
+import android.content.Intent
+import android.content.pm.PackageInfo
+import android.content.pm.PackageManager
+import android.content.pm.ResolveInfo
+import android.os.Build
+
+fun PackageManager.queryIntentActivitiesCompat(intent: Intent, flags: Int): List<ResolveInfo> =
```
paetz, line 26: What is a "compat"?
Would it make sense to call it `queryCustomIntentActivities()` or `queryIntentActivities_patch()` to make clear that we've patched the actual package manager?
Same for the other function below.

paetz (approved):

app/src/main/java/com/yacgroup/yacguide/extensions/PackageManager.kt
chrgernoe: "compat" stands for compatibility. I think it is also used by Google for a set of libraries `AppCompat` which are compatible with older versions.

PR #413, opened 2023-03-03
Show what's new dialog even if default sector was set

chrgernoe:


app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
```diff
@@ -62,12 +62,15 @@ class LaunchActivity : AppCompatActivity() {
...
+                putExtra(IntentConstants.SHOW_WHATS_NEW, true)
+            }
         else
             Intent(this, SectorActivity::class.java).apply {
                 putExtra(IntentConstants.CLIMBING_OBJECT_LEVEL, ClimbingObjectLevel.eSector.value)
                 putExtra(IntentConstants.CLIMBING_OBJECT_PARENT_ID, defaultRegionId)
                 putExtra(IntentConstants.CLIMBING_OBJECT_PARENT_NAME, _db.getRegion(defaultRegionId)?.name.orEmpty())
+                putExtra(IntentConstants.SHOW_WHATS_NEW, true)
             }
         startActivity(intent)
```
paetz, line 75: Why not checking for versionUpdate and showing dialog here before switching to Country- or SectorActivity? Would reduce a lot of code and duplication.

paetz (changes requested):

app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
chrgernoe: That doesn't work. I tried this. The problem is that the dialog cannot block the activity method. Hence the dialog pops up but gets closed immediately because the switch to database takes place. 
I tried to control the switch to the database view via the dialog but it gets very complicated for the case that the user want to see the news. In this case another activity is started.
Any other solution is welcome!

paetz: Ok I see. Then let's implement a compromise:
The `checkForVersionUpdate()` is done here in `LaunchActivity` and its result is passed as the value of `IntentConstants.SHOW_WHATS_NEW`, so that you do not have to check for versionUpdate in Country- and SectorActivity.

paetz (dismissed):

PR #422, opened 2023-04-12
Remove project specific Docker environment

chrgernoe:
Since GitHub workflows/actions implementation advanced a lot since I implemented the whole Docker stuff, I decided to simply things, get rid of the project specific Docker environment and use the GitHub flow instead.


paetz (approved):
Reducing code is always nice :)

Is CI now running unit and instrumentation tests?

chrgernoe:
> Reducing code is always nice :)
> 
> Is CI now running unit and instrumentation tests?

Yes, the CI runs both tests but did it also before. Code coverage comes next.

PR #430, opened 2023-06-11
Add clear text icon to search fields

### LINKED ISSUE #425: Clearing-cross for search field
When searching an object via the search box on the rock and route list, a clearing cross should appear at the end of the search box to remove everything that has been typed into the box yet.

We should use the Andoid SearchView widget, which provides this function already: https://developer.android.com/reference/android/widget/SearchView

chrgernoe:
Fixes #425

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/PartnersActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/PartnersActivity.kt
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2019, 2022, 2023 Axel Paetzold
+ * Copyright (C) 2023 Christian Sommer
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -28,6 +29,7 @@ import android.text.TextWatcher
 import android.util.SparseIntArray
 import android.view.MenuItem
 import android.view.View
+import android.view.View.INVISIBLE
 import android.view.inputmethod.InputMethodManager
 import android.widget.EditText
 import android.widget.Toast
@@ -39,6 +41,7 @@ import com.yacgroup.yacguide.database.Partner
 import com.yacgroup.yacguide.list_adapters.*
 import com.yacgroup.yacguide.utils.DialogWidgetBuilder
 import com.yacgroup.yacguide.utils.IntentConstants
+import com.yacgroup.yacguide.utils.SearchBarHandler
 import com.yacgroup.yacguide.utils.VisualUtils
 
 import java.util.ArrayList
@@ -46,6 +49,7 @@ import java.util.ArrayList
 class PartnersActivity : AppCompatActivity() {
 
     private val _ascendPartnerCount = SparseIntArray()
+    private lateinit var _searchBarHandler: SearchBarHandler
     private lateinit var _viewAdapter: ListViewAdapter<Partner>
     private lateinit var _visualUtils: VisualUtils
     private lateinit var _db: DatabaseWrapper
@@ -74,6 +78,12 @@ class PartnersActivity : AppCompatActivity() {
                 _displayContent()
             }
         })
+
+        _searchBarHandler = SearchBarHandler(
+            searchBarLayout = findViewById(R.id.searchBarLayout),
+            checkBoxVisibility = INVISIBLE,
+            searchHintResource = R.string.partner_search,
+        )
 
         _viewAdapter = ListViewAdapter(ItemDiffCallback(
             _areItemsTheSame = { partner1, partner2 -> partner1.id == partner2.id },
--- a/app/src/main/java/com/yacgroup/yacguide/PartnersActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/PartnersActivity.kt
@@ -29,7 +29,6 @@ import android.text.TextWatcher
 import android.util.SparseIntArray
 import android.view.MenuItem
 import android.view.View
-import android.view.View.INVISIBLE
 import android.view.inputmethod.InputMethodManager
 import android.widget.EditText
 import android.widget.Toast
@@ -81,7 +80,7 @@ class PartnersActivity : AppCompatActivity() {
 
         _searchBarHandler = SearchBarHandler(
             searchBarLayout = findViewById(R.id.searchBarLayout),
-            checkBoxVisibility = INVISIBLE,
+            checkBoxVisibility = View.INVISIBLE,
             searchHintResource = R.string.partner_search,
         )
 
--- a/app/src/main/java/com/yacgroup/yacguide/utils/SearchBarHandler.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/utils/SearchBarHandler.kt
@@ -26,6 +26,7 @@ import android.view.inputmethod.InputMethodManager
 import android.widget.CheckBox
 import android.widget.EditText
 import androidx.constraintlayout.widget.ConstraintLayout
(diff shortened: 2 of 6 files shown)


app/src/main/java/com/yacgroup/yacguide/utils/SearchBarHandler.kt
```diff
@@ -52,8 +53,10 @@ class SearchBarHandler(searchBarLayout: ConstraintLayout,
             })
         }
 
-        searchBarLayout.findViewById<EditText>(R.id.searchEditText).apply {
+        searchBarLayout.findViewById<TextInputLayout>(R.id.searchTextInputLayout).apply {
             setHint(searchHintResource)
+        }
+        searchBarLayout.findViewById<EditText>(R.id.searchEditText).apply {
```
paetz, line 59: You have changed the widget's type in the xml to `com.google.android.material.textfield.TextInputEditText`. Does this still correspond to Kotlin's `EditText` type?

app/src/main/res/layout/activity_partners.xml
```diff
@@ -22,26 +22,37 @@
     android:layout_width="match_parent"
     android:layout_height="match_parent">
 
+    <com.google.android.material.textfield.TextInputLayout
```
paetz, line 25: Why did you move the TextField out of the `LinearLayout`?  Do we actually still need the `LinearLayout` below (It only contains one element in your solution)?

app/src/main/res/layout/searchbar.xml
```diff
@@ -27,21 +28,32 @@
...
-        android:layout_height="45dp"
-        android:layout_marginTop="8dp"
-        android:maxLines="1"
-        android:singleLine="true"
-        app:layout_constraintRight_toRightOf="parent"
         app:layout_constraintTop_toTopOf="parent" />
 
-</androidx.constraintlayout.widget.ConstraintLayout>
\ No newline at end of file
+    <com.google.android.material.textfield.TextInputLayout
```
paetz, line 38: Maybe it makes sense for `PartnersActivity` to include this searchbar and make the checkbox above invisible at the activity's start-up instead of defining the same component on its own? Doing so, we could reduce quite some code-duplication.

paetz: I could even think of re-using the checkbox for PartnerActivity e.g. for sth. like "sort by name" to either sort partners by name or number of shared routes.

paetz (dismissed):

app/src/main/res/layout/searchbar.xml
chrgernoe: I thought about this, too but I didn't want to change too many things. Anyway, I will include this change into the the PR.

app/src/main/res/layout/activity_partners.xml
chrgernoe: Because I need the layout constraint feature for correct placement of the search field. The `LinearLayout` acts as container of the included XML so that the list view can be aligned correctly in the current layout. Since the included XML is used in different layouts, I probably break some other layout. Another PR is required, to unify all the layouts where the list view is included.

app/src/main/java/com/yacgroup/yacguide/utils/SearchBarHandler.kt
chrgernoe: It's a sub-class of `EditText` (https://developer.android.com/reference/com/google/android/material/textfield/TextInputEditText) which shall be used within the `TextInputLayout`.

app/src/main/java/com/yacgroup/yacguide/PartnersActivity.kt
```diff
@@ -75,6 +79,12 @@ class PartnersActivity : AppCompatActivity() {
             }
         })
 
+        _searchBarHandler = SearchBarHandler(
+            searchBarLayout = findViewById(R.id.searchBarLayout),
+            checkBoxVisibility = INVISIBLE,
```
paetz, line 84: Please use `View.INVISIBLE` and remove the `android.view.View.INVISIBLE` import

app/src/main/java/com/yacgroup/yacguide/utils/SearchBarHandler.kt
```diff
@@ -22,38 +23,45 @@ import android.content.SharedPreferences
...
-                       checkBoxDefaultValue: Boolean,
-                       private val _settings: SharedPreferences,
-                       private val _initCallback: (Boolean) -> Unit,
-                       private val _updateCallback: (String, Boolean) -> Unit) {
+                       checkBoxTitle: String = "",
+                       checkBoxDefaultValue: Boolean = false,
+                       private val _settings: SharedPreferences? = null,
+                       private val _initCallback: ((Boolean) -> Unit)? = null,
+                       private val _updateCallback: ((String, Boolean) -> Unit)? = null,
+                       checkBoxVisibility: Int = VISIBLE) {
```
paetz, line 41: - [ ] Please use `View.INVISIBLE` and remove the `android.view.View.INVISIBLE` import
- [ ] Please put this parameter 3 lines above (behind checkBoxDefaultValue)

app/src/main/java/com/yacgroup/yacguide/utils/SearchBarHandler.kt
```diff
@@ -22,38 +23,45 @@ import android.content.SharedPreferences
...
                        searchHintResource: Int,
-                       checkBoxTitle: String,
-                       checkBoxDefaultValue: Boolean,
-                       private val _settings: SharedPreferences,
-                       private val _initCallback: (Boolean) -> Unit,
-                       private val _updateCallback: (String, Boolean) -> Unit) {
+                       checkBoxTitle: String = "",
+                       checkBoxDefaultValue: Boolean = false,
+                       private val _settings: SharedPreferences? = null,
+                       private val _initCallback: ((Boolean) -> Unit)? = null,
```
paetz, line 39: Wouldn't it be safer (and more convenient) to default `_initCallback` and `_updateCallback` to empty lambdas?
--> `_initCallback: (Boolean) -> Unit = {}`

app/src/main/res/layout/activity_partners.xml
```diff
@@ -25,23 +26,16 @@
     <LinearLayout
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:orientation="vertical"
         android:clickable="true"
         android:focusable="true"
         android:focusableInTouchMode="true"
+        android:orientation="vertical"
+        app:layout_constraintEnd_toEndOf="parent"
```
paetz, line 33: Why is this necessary?

paetz (dismissed):

app/src/main/res/layout/activity_partners.xml
chrgernoe: You mean the `LinearLayout`? I answered the question already above?

app/src/main/java/com/yacgroup/yacguide/utils/SearchBarHandler.kt
chrgernoe: Because the class is used without named arguments in the code, I wanted to make it compatible. But anyway, I will change those lines, too.

app/src/main/res/layout/activity_partners.xml
paetz: I mean `app:layout_constraintEnd_toEndOf="parent"`

chrgernoe: You are right, this is not necessary. I going to remove it.

PR #434, opened 2023-08-10
Enable stacked bar charts for statistics

### LINKED ISSUE #433: Enable stacked bar charts for statistics

paetz:
Resolves #433

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/list_adapters/StatisticsViewAdapter.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/list_adapters/StatisticsViewAdapter.kt
@@ -17,6 +17,7 @@
 
 package com.yacgroup.yacguide.list_adapters
 
+import android.graphics.Color
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/yacgroup/yacguide/list_adapters/StatisticsViewAdapter.kt
```diff
@@ -61,6 +101,17 @@ class StatisticsViewAdapter : ListAdapter<Statistic, RecyclerView.ViewHolder>(St
...
         (viewHolder as StatisticViewHolder).bind(getItem(position))
     }
+
+    private fun _generateDataSet(stat: Statistic): BarDataSet {
+        val colorOffset = 155 / stat.stackTitles.size
+        return BarDataSet(stat.data.values.toList(), stat.name).apply {
+            label = ""
+            stackLabels = stat.stackTitles.toTypedArray()
+            colors = List(stat.stackTitles.size) { idx ->
+                Color.rgb(100, 255 - idx * colorOffset, 100)
```
chrgernoe, line 111: The color definition should be derived from a consisten app color defined in the color resource XML file.

chrgernoe (approved):

app/src/main/java/com/yacgroup/yacguide/list_adapters/StatisticsViewAdapter.kt
paetz: Yes but how? The colors are depending on the number of stacks, so their values are not absolute.

chrgernoe: The base color is red=100, green=255 and blue=100, which can be defined as hex code in the resource file. In the source code you just extract the RGB values via some function. However, for larger numbers of data points, the resulting color gets grey anyway. To avoid this, you need to adapt the offset, so that you don't leave a certain color range.

BTW: What happens if you `idx` exceeds 255 at this point?

paetz: Since `colorOffset` is derived from `stackTitles` size, `idx * offset` will never exceed 255.

PR #438, opened 2023-08-29
Add ascends-per-year count statistic

### LINKED ISSUE #244: Add statistic: Number of ascends per year
- Add a new statistic to the statistics view like "Anzahl Begehungen"
- Create a stacked bar chart with the number of ascends (lead and follow separated) per year 

Note: Since the number of years can be huge, I suggest a vertical bar chart.

paetz:
Resolves #244

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/StatisticsActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/StatisticsActivity.kt
@@ -63,20 +63,27 @@ class StatisticsActivity : BaseNavigationActivity() {
             Statistic(
                 getString(R.string.ascends_per_year),
                 listOf(
-                    getString(R.string.others),
+                    getString(R.string.follow) + "/" + getString(R.string.others),
                     getString(R.string.lead),
                     getString(R.string.solo)
                 ),
-                soloCounts.mapIndexed { idx, ascendCount ->
+                otherCounts.mapIndexed { idx, ascendCount ->
                     val yearLabel = if (ascendCount.year == 0) "" else ascendCount.year.toString()
-                    yearLabel to BarEntry(idx.toFloat(), floatArrayOf(
-                        otherCounts[idx].count.toFloat(),
-                        leadCounts[idx].count.toFloat(),
-                        soloCounts[idx].count.toFloat()
-                    ))
+                    try {
+                        yearLabel to BarEntry(idx.toFloat(), floatArrayOf(
+                            otherCounts[idx].count.toFloat(),
+                            leadCounts[idx].count.toFloat(),
+                            soloCounts[idx].count.toFloat()
+                        ))
+                    } catch (e: IndexOutOfBoundsException) {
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/yacgroup/yacguide/StatisticsActivity.kt
```diff
@@ -48,21 +48,34 @@ class StatisticsActivity : BaseNavigationActivity() {
     }
 
     private fun _displayContent() {
-        // Placeholders:
+        // The following counts contain all years, even if there was no ascent of the according style.
+        // This means that the list sizes are all equal.
```
paetz, line 52: There should be some assertion in place that all those counts indeed contain the same years and are of the same length.

app/src/main/java/com/yacgroup/yacguide/StatisticsActivity.kt
```diff
@@ -48,21 +48,34 @@ class StatisticsActivity : BaseNavigationActivity() {
...
+        val soloCounts = _db.getAscendCountPerYear(AscendStyle.eSOLO.id)
+        val leadCounts = _db.getAscendCountPerYear(AscendStyle.eONSIGHT.id, AscendStyle.eHOCHGESCHLEUDERT.id)
+        val followCounts = _db.getAscendCountPerYear(AscendStyle.eALTERNATINGLEADS.id, AscendStyle.eHINTERHERGEHAMPELT.id)
+        val unknownCounts = _db.getAscendCountPerYear(AscendStyle.eUNKNOWN.id)
+        val otherCounts = followCounts.map { followCount ->
```
chrgernoe, line 57: When I first looked at the statistic, I was wondering why there is no "follow" bar. I would like to separate "unknown" and "follow" bars. Actually, there should no "unknown" bar at all.

app/src/main/java/com/yacgroup/yacguide/StatisticsActivity.kt
chrgernoe: How can this happen?

paetz: It should not happen because of the way the SQL queries are constructed. However, as a user of this API, it's not really obvious that these lists are all of the same length with the same keys (year).

app/src/main/java/com/yacgroup/yacguide/StatisticsActivity.kt
paetz: But how to handle style "Alternating leads" then? (is now part of "others")

chrgernoe: As discussed offline, it might get to messy to have another style in the diagram. However, "Sonstige" is for me still a bit confusing. "Nachstieg, etc." or so could be an alternative that is a bit more obvious, because "Nachstieg" is one of the most used ascend styles.

chrgernoe (approved):

PR #444, opened 2023-12-09
Added ascends-per-partner statistic

### LINKED ISSUE #439: Add statistic: Most frequent climbing partners

paetz:
Resolves #439


chrgernoe (changes requested):
For some reason only every second partner is shown in the y-axis on my device (Pixel 4a, Android 13) even though it is enough space. Only after zooming it is possible to see the partner name, which is very inconvenient,

PR #445, opened 2023-12-27
Update of SDK, libraries and tool chain

chrgernoe:


paetz:
- [ ] Are these new files in the .idea directory intended to be committed? Or should we add them to .gitignore?

PR #448, opened 2024-01-14
Show creation date of comments

### LINKED ISSUE #447: Show  creation date of comments
A user requested to include the creation date of a comment to figure out whether content is still valid or not.

Implementation proposal: Include the date the square brackets of the comment author.

chrgernoe:
Fixes #447

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/network/SectorParser.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/network/SectorParser.kt
@@ -28,6 +28,7 @@ import com.yacgroup.yacguide.utils.ParserUtils
 import com.yacgroup.yacguide.utils.DateUtils
 import org.json.JSONArray
 import org.json.JSONException
+import org.json.JSONObject
 import java.lang.RuntimeException
 import java.util.*
 import kotlin.jvm.Throws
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/yacgroup/yacguide/network/SectorParser.kt
```diff
@@ -136,10 +137,11 @@ class SectorParser(private val _db: DatabaseWrapper) : JSONWebParser() {
         for (i in 0 until jsonComments.length()) {
             val jsonComment = jsonComments.getJSONObject(i)
             val user = jsonComment.getString("username")
+            val date = _parseCommentDate(jsonComment.getString("datum"))
```
paetz, line 140: There is also an "adatum" field which may contain the date of the latest change. Would this make more sense?

app/src/main/java/com/yacgroup/yacguide/network/SectorParser.kt
```diff
@@ -280,4 +283,10 @@ class SectorParser(private val _db: DatabaseWrapper) : JSONWebParser() {
         }
         _db.updateRocks(updatedRocks)
     }
+
+    /**
+     * Parse comment date-times in the format yyyy-MM-dd HH:mm:ss
+     * and return only the date in the format yyyy-MM-dd.
+     */
+    private fun _parseCommentDate(date: String): String = DateUtils.formatDate(date.split(" ")[0])
```
paetz, line 291: - [ ] You should add error handling here in case `date` is an empty string.

paetz (dismissed):

app/src/main/java/com/yacgroup/yacguide/network/SectorParser.kt
chrgernoe: I had a look into the PHP code of [sandsteinklettern.de](sandsteinklettern.de) and for my understanding it is the archive date. It is updated, when the comment gets updated.

chrgernoe: I'm now prioritizing the `adatum` and use `datum` as fallback.

paetz (approved):

Issue #429, opened 2023-06-04
Use Android Preference Library for App Settings


chrgernoe:
Instead of writing the settings activity by hand, we should use the Android preference library. 

https://developer.android.com/develop/ui/views/components/settings

PR #454, opened 2024-02-18
Fix that shared preferences use description strings instead of unique keys

### LINKED ISSUE #453: Shared preferences uses description strings instead of unique keys
Currently some preferences are stored using the the UI description strings instead of defined keys. That means, if the strings are changed for some reasons, the settings are lost.

**Proposal:**
Use defined keys and implement code that migrates existing settings to the new structure.

chrgernoe:
* Fixes #453 
* Needs merge of #429 before.

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
@@ -22,12 +22,12 @@ import android.content.Intent
 import android.content.SharedPreferences
 import android.os.Bundle
 import android.os.CountDownTimer
-import android.util.Log
 import android.widget.ImageView
 import androidx.appcompat.app.AppCompatActivity
 import com.yacgroup.yacguide.database.DatabaseWrapper
 import com.yacgroup.yacguide.extensions.getPackageInfoCompat
 import com.yacgroup.yacguide.extensions.versionCodeCompat
+import com.yacgroup.yacguide.migration.PreferencesMigration
 import com.yacgroup.yacguide.network.CountryAndRegionParser
 import com.yacgroup.yacguide.utils.IntentConstants
 
@@ -53,12 +53,13 @@ class LaunchActivity : AppCompatActivity() {
 
         _db = DatabaseWrapper(this)
         _customSettings = getSharedPreferences(getString(R.string.preferences_filename), Context.MODE_PRIVATE)
-        _checkForSettingsMigration()
+        _checkForPreferencesMigration()
 
         _updateHandler = UpdateHandler(this, CountryAndRegionParser(_db))
         _updateHandler.update(
             onUpdateFinished = { findViewById<ImageView>(R.id.iconImageView).setImageResource(R.drawable.ic_start_done) },
-            isSilent = true)
+            isSilent = true
+        )
         Timer().start()
     }
 
@@ -86,7 +87,8 @@ class LaunchActivity : AppCompatActivity() {
      */
     private fun _checkForVersionUpdate(): Boolean {
         val savedVerCode = _customSettings.getInt(
-            getString(R.string.pref_key_version_code), 0)
+            getString(R.string.pref_key_version_code), 0
+        )
         val curVerCode = try {
             packageManager.getPackageInfoCompat(packageName, 0).versionCodeCompat
         } catch (e: Exception) {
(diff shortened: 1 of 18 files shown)


chrgernoe:
We should also make the naming in the code consistent. Currently there are two different wordings:
* settings
* preferences

The Android library uses the the word _preferences_.

app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
```diff
@@ -84,19 +86,92 @@ class LaunchActivity : AppCompatActivity() {
...
+                    Pair(R.string.pref_key_color_lead, R.color.pref_default_color_lead),
+            R.string.follow to
+                    Pair(R.string.pref_key_color_follow, R.color.pref_default_color_follow)
+        )
+        _customSettings.let {
+            it.edit().apply {
+                for ((oldResource, newResourcePair) in settingsResourceMap) {
+                    val oldKey = resources.getString(oldResource)
+                    val newKey = resources.getString(newResourcePair.first)
+                    Log.d("[PREF]", "Checking for deprecated key '$oldKey'.")
```
paetz, line 153: I think, if we want to introduce logging, we should check for creating a real logger module or so. So I'm not sure if we should really introduce it as part of this PR.

app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
```diff
@@ -84,19 +86,92 @@ class LaunchActivity : AppCompatActivity() {
...
         } else {
             false
         }
     }
+
+    /**
+     * Check whether the settings need to be migrated or not.
+     */
+    private fun _checkForSettingsMigration() {
+        if (_customSettings.getInt(getString(R.string.pref_key_pref_version), 1) == 1) {
```
paetz, line 109: Wouldn't it make more sense to check for the latest version? Otherwise, we would always need to remember adapting this line when we increase the version.

paetz, line 109: Wouldn't it make more sense to check for the latest version? Otherwise, we would always need to remember adapting this line when we increase the version.

paetz, line 109: Wouldn't it make more sense to check for the latest version? Otherwise, we would always need to remember adapting this line when we increase the version.

app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
```diff
@@ -84,19 +86,92 @@ class LaunchActivity : AppCompatActivity() {
...
+            _migrateSettings()
+        }
+    }
+
+    /**
+     * Migrate settings, if necessary, to fix issue https://github.com/YacGroup/yacguide/issues/453.
+     */
+    private fun _migrateSettings() {
```
paetz, line 117: - [ ] All this migration stuff should not be part of `LaunchActivity`. Please move this functionality to a separate file.

app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
```diff
@@ -84,19 +86,92 @@ class LaunchActivity : AppCompatActivity() {
...
+        )
+        _customSettings.let {
+            it.edit().apply {
+                for ((oldResource, newResourcePair) in settingsResourceMap) {
+                    val oldKey = resources.getString(oldResource)
+                    val newKey = resources.getString(newResourcePair.first)
+                    Log.d("[PREF]", "Checking for deprecated key '$oldKey'.")
+                    if (it.contains(oldKey)) {
+                        Log.d("[PREF]", "Migrating key '$oldKey' to '$newKey'.")
+                        if (oldResource in listOf(R.string.lead, R.string.follow)) {
```
paetz, line 156: Creating a new list by using `listOf` performs worse than `(oldResource == R.string.lead || oldResource == R.string.follow)`

app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
paetz, line 117: This migration stuff is quite critical. I would suggest to have a separate, unit-tested class for this.

app/src/main/java/com/yacgroup/yacguide/utils/SearchBarHandler.kt
```diff
@@ -35,6 +35,7 @@ class SearchBarHandler(searchBarLayout: ConstraintLayout,
                        private val _checkBoxTitle: String = "",
```
paetz, line 35: I think, this can be a normal argument now, since it is only used in `init {}`?

paetz (changes requested):

paetz:
This is a quite critical change. I suggest to trigger a dev release after merging it to have some more time for testing.

app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
chrgernoe: OK, I agree. It was mainly for debugging but I will remove it.

app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
chrgernoe: Duplicated comment.

app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
chrgernoe: Makes sense. I'm going to implement a test for this.

app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
chrgernoe: Duplicated comment.

paetz (approved):

PR #471, opened 2025-01-03
Add new statistic: Newly collected rocks per year

### LINKED ISSUE #464: Add statistic: Number of newly collected rocks per year
The statistic should show the number of rocks per year that have been collected.
Take into account what type of rock the user is collecting in the SharedPreferences.

paetz:
Resolves #464 .

To improve performance, this feature has been implemented into one single sql request.

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/statistics/StatisticGenerator.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/statistics/StatisticGenerator.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 Axel Paetzold
+ * Copyright (C) 2023, 2025 Axel Paetzold
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/yacgroup/yacguide/statistics/StatisticGenerator.kt
```diff
@@ -64,15 +67,38 @@ class StatisticGenerator(private val _ctx: Context) {
...
+        val endStyleId = if (considerOnlyLeads)
+                AscendStyle.eHOCHGESCHLEUDERT.id
+            else
+                AscendStyle.eHINTERHERGEHAMPELT.id
+        val rockCounter = RockCounter(RockCounterConfig.generate(_ctx, _customSettings))
+        val statsTitle = _ctx.getString(R.string.newly_collected_rocks_per_year)
+        return Statistic(
+            statsTitle,
+            listOf(statsTitle),
+            _db.getNewlyAscendedRockCountsPerYear(startStyleId, endStyleId, rockCounter.getConsideredRockTypes(), rockCounter.getExcludedRockStates()).associate { rockCount ->
```
chrgernoe, line 97: I see enums as something where the programmer does not need to care about the value behind. With this `startStyleId`/`endStyleId` approach this is not true anymore. I think, we also have other places in the code like this. Maybe we should consider to handle names groups of styles which are handled on a central place.

chrgernoe (approved):

app/src/main/java/com/yacgroup/yacguide/statistics/StatisticGenerator.kt
paetz: I agree, some refactoring in the future might be neat.

PR #472, opened 2025-01-07
Show usage info for certain activities

### LINKED ISSUE #460: Add info dialogs when entering certain activities
E.g. when entering the "Gebiete verwalten" activity, a user does know that updating/removing is done via swiping.

Add a dialog for those kinds of hidden features with a checkbox like "Nicht mehr anzeigen". The latter may be stored in the SharedPreferences.

paetz:
Resolves #460


chrgernoe:
As discussed offline, the `Snackbar` design element is in my eyes not the correct way to do this. Maybe we should use a library like this: https://github.com/KeepSafe/TapTargetView

paetz:
> As discussed offline, the `Snackbar` design element is in my eyes not the correct way to do this. Maybe we should use a library like this: https://github.com/KeepSafe/TapTargetView

For this tiny feature, I would not integrate another 3rd party lib. I guess, we would need to use our own pop-up dialog for this.

chrgernoe (approved):

PR #474, opened 2025-01-08
Add hint for user to retry update on server error

### LINKED ISSUE #473: Add "Retry" info to error dialog on updating regions
Sometimes we observe server errors when updating a region. In this case, we tell the user via dialog that something went wrong.
Since there is not solution for this issue, the user can only retry.

Add a notification to the error dialog that the user should retry the update mechanism for the failed regions.

paetz:
Resolves #473 

Misusing the comment dialog for this by adding another text field yields some problems, so I have decided to simply put it into the according Toast.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -89,7 +89,7 @@
     <!-- Toasts -->
     <string name="refresh_successful">Aktualisierung erfolgreich</string>
     <string name="region_deleted">Gebiet gelöscht</string>
-    <string name="refresh_failed">Aktualisierung fehlgeschlagen</string>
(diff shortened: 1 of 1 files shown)


chrgernoe:
Instead of using the comment dialog, the Android Snackbar element would something that offers the possibility to retry.

chrgernoe (approved):
Instead of using the comment dialog, the Android Snackbar element would something that offers the possibility to retry.

paetz:
> Instead of using the comment dialog, the Android Snackbar element would something that offers the possibility to retry.

Having a toast for "Aktualisierung fehlgeschlagen" and a snackbar for a hint to retry at the same time is not nice. We could maybe put both into a snackbar but I don't see much difference between displaying it via toast or snackbar.

PR #479, opened 2025-02-05
Use SDK version 35

chrgernoe:


chrgernoe:
> What about changing CI job to use API level 35 for running Instrumented tests? 

It is part of the PR.

PR #480, opened 2025-02-24
Update chart library to be able to configure axis label count

paetz:
This PR fixes the issue that we could only have up to 25 axis labels in our statistic graphs.

I have found an actively maintained fork of our chart library that allows configuration of that limit, so I have replaced the original lib by this fork and configured the upper limit of axis labels before displaying it.


paetz:
Do we need to list the new dependency somewhere or is this done automatically?

chrgernoe (approved):

chrgernoe:
> Do we need to list the new dependency somewhere or is this done automatically?

That happens automatically.

PR #481, opened 2025-03-18
Remove some non-determinism

licaon-kter:
Your app is not built reproducible in F-Droid, but we continuously test older versions on https://verification.f-droid.org and would like more and more apps to become repro.

Looking at this report: https://verification.f-droid.org/com.yacgroup.yacguide_9000.apk.diffoscope.html

We can fix that based on: https://github.com/mikepenz/AboutLibraries/issues/784#issuecomment-1205501876

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -121,6 +121,7 @@ android {
 aboutLibraries {
     prettyPrint = true
     configPath = 'config/aboutLibraries'
+    excludeFields = arrayOf("generated")
 }
 
 dependencies {


app/build.gradle
```diff
@@ -121,6 +121,7 @@ android {
 aboutLibraries {
     prettyPrint = true
     configPath = 'config/aboutLibraries'
+    excludeFields = arrayOf("generated")
```
chrgernoe, line 124: @licaon-kter: This is the wrong syntax because our `build.gradle` file uses Groovy and not Kotlin. The correct syntax should be `excludeFields = ['generated']`.

chrgernoe (changes requested):

PR #498, opened 2026-01-11
Allow manual date input

### LINKED ISSUE #497: Allow manual date input
The user wants to input the tour date without using the date picker. Especially, if you input many entries this is much faster.

chrgernoe:
Fixes #497 

* The initial default is the calendar mode.
* The user can switch the mode in the widget. The choice is remembered everything the date is confirmed via OK.
* In the  text mode, the default date is empty for new tourbook entries to give the user the complete freedom without editing/clearing the date.
* The date format is checked by the widget.

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
@@ -133,22 +133,21 @@ class AscendActivity : AppCompatActivity() {
             resources.getInteger(R.integer.pref_default_calendar_input_mode)
         )
         val ascendDatePickerUtils = AscendDatePickerUtils(_ascend)
-        val datePicker = MaterialDatePicker.Builder.datePicker()
+        MaterialDatePicker.Builder.datePicker()
             .setTheme(R.style.ThemeOverlay_App_MaterialCalendar)
             .setSelection(ascendDatePickerUtils.getSelection(
                 AscendDatePickerInputMode.from(calendarInputMode))
             )
             .setInputMode(calendarInputMode)
-            .build()
-        datePicker.addOnPositiveButtonClickListener { selection ->
-            ascendDatePickerUtils.updateAscend(selection)
-            _activityViewBinding.dateEditText.setText("${_ascend.day}.${_ascend.month}.${_ascend.year}")
-            // Save last selected input mode in preferences
-            _customSettings.edit {
-                putInt(getString(R.string.pref_key_calendar_input_mode), datePicker.inputMode)
-            }
+            .build().let { datePicker ->
+                datePicker.addOnPositiveButtonClickListener { selection ->
+                    ascendDatePickerUtils.updateAscend(selection)
+                    _activityViewBinding.dateEditText.setText("${_ascend.day}.${_ascend.month}.${_ascend.year}")
+                    // Save last selected input mode in preferences
+                    _customSettings.edit { putInt(getString(R.string.pref_key_calendar_input_mode), datePicker.inputMode) }
+                }
+                datePicker.show(supportFragmentManager, "MATERIAL_DATE_PICKER")
         }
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/setup.xml
```diff
@@ -44,7 +44,10 @@
 
     <!-- Preference keys -->
     <string name="pref_key_pref_version">preference_version</string>
-    <integer name="pref_version">2</integer>
+    <integer name="pref_version">3</integer>
```
chrgernoe: With this increase of the version, the migration test from `v1` to `v2`, introduced due to issue #453, fails. We have two options here:
1. Make the tests compatible with version 3. Basically that asumes that a user has installed a app that still uses `v1` and suddenly installs the most recent one but never went through the migration process.
2. Remove the migration code plus the test completely, which assumes that all users are now migrated.

paetz: I don't get it. We should actually have two migrations now: v1 -> v2 and v2 -> v3. I would have assumed that defining both migrations should still make v1->v2- Tests pass?

app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
```diff
@@ -52,6 +54,7 @@ class AscendActivity : AppCompatActivity() {
     private lateinit var _partnerResultLauncher: ActivityResultLauncher<Intent>
     private var _outdatedAscend: Ascend? = null
     private var _route: Route? = null
+    private lateinit var _customSettings: SharedPreferences
```
paetz: I'd find it nicer to place this declaration below the other `lateinit` var in line 54

paetz (changes requested):

app/src/main/res/values/setup.xml
chrgernoe: There is no migration from v2 to v3 required because the date picker adds the correct settings key. The migration from v1 to v2  was only necessary because there was a bug.

app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
```diff
@@ -122,17 +129,49 @@ class AscendActivity : AppCompatActivity() {
...
+        } else {
+            if (calendarInputMode == 0) {
+                MaterialDatePicker.todayInUtcMilliseconds()
+            } else {
+                // Do not fill the calendar with the current date in text mode
+                // because this mode is intended for fast input of older dates.
+                null
+            }
+        }
+        val datePicker = MaterialDatePicker.Builder.datePicker()
```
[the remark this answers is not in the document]
chrgernoe: There was no custom DatePickerDialog. It came from an android library.

app/src/main/java/com/yacgroup/yacguide/LaunchActivity.kt
```diff
@@ -116,6 +118,7 @@ class LaunchActivity : AppCompatActivity() {
         )
         if (_customSettings.getInt(getString(R.string.pref_key_pref_version), 1) == 1) {
```
paetz: Don't we need an `else if (_customSettings.getInt(getString(R.string.pref_key_pref_version), 1) == 2)` now?
Btw: Is the default value in `getInt()` (= 1) correct? This would apply if the pref does not exist right? And this is only the case for pref-version 1? I am just curious

app/src/main/res/values/setup.xml
```diff
@@ -44,7 +44,10 @@
...
+
+    <string name="pref_key_calendar_input_mode">calendar_input_mode</string>
+    <integer name="pref_default_calendar_input_mode">0</integer>
```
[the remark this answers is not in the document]
chrgernoe: I introduced the enum but since here I cannot use the enum, I have to hard-code the number but left a comment.

PR #504, opened 2026-03-18
Fix app crash when display is turned to landscape format during tourbook edit or input

### LINKED ISSUE #503: App crashes when display is turned to landscape format during tourbook edit or entry

chrgernoe:
* Fixes actual null pointer exception
* Fix/Feature: Preserve user inputs when display mode is changed

Fixes #503

### DIFF
--- a/app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
+++ b/app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
@@ -72,16 +72,19 @@ class AscendActivity : BaseActivity<ActivityAscendBinding>() {
         _route = _db.getRoute(_outdatedAscend?.routeId
                     ?: intent.getIntExtra(IntentConstants.CLIMBING_OBJECT_PARENT_ID, DatabaseWrapper.INVALID_ID))
         // Beware: _route may still be null (if the route of this ascend has been deleted meanwhile)
-        _ascend = Ascend(
-            id = _outdatedAscend?.id ?: 0,
-            routeId = _outdatedAscend?.routeId ?: _route!!.id,
-            styleId = _outdatedAscend?.styleId ?: 0,
-            year = _outdatedAscend?.year ?: 0,
-            month = _outdatedAscend?.month ?: 0,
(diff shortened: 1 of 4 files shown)


paetz (changes requested):

app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
```diff
@@ -72,16 +72,19 @@ class AscendActivity : BaseActivity<ActivityAscendBinding>() {
...
-            styleId = _outdatedAscend?.styleId ?: 0,
-            year = _outdatedAscend?.year ?: 0,
-            month = _outdatedAscend?.month ?: 0,
-            day = _outdatedAscend?.day ?: 0,
-            partnerIds = _outdatedAscend?.partnerIds ?: ArrayList(),
-            notes = _outdatedAscend?.notes ?: ""
-        )
+        _ascend = savedInstanceState.let {
+            Ascend(
+                id = it?.getInt("ascend_id") ?: _outdatedAscend?.id ?: 0,
```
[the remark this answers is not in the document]
chrgernoe: Thx for the feedback. With your trigger, I found a even more elegant implementation.

app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
```diff
@@ -174,7 +186,7 @@ class AscendActivity : BaseActivity<ActivityAscendBinding>() {
             onItemSelectedListener = object : AdapterView.OnItemSelectedListener {
                 override fun onItemSelected(
                     parent: AdapterView<*>,
-                    view: View,
+                    view: View?,
```
paetz, line 189: Why is this change necessary?

paetz (approved):

app/src/main/java/com/yacgroup/yacguide/AscendActivity.kt
chrgernoe: That was the actual reason of the null pointer exception. The `view` can be null, if the display is turned.

PR #513, opened 2026-09-17
Fix empty CSV/verbose JSON tourbook export in release builds

### LINKED ISSUE #512: Name-based tourbook export results in empty file
The export of the tourbook as a name-based CSV or JSON results in an empty file.

**Root cause**

The issue is caused by the app optimization `R8` introduced with `v0.9.4`@ commit 1d44f9a5667f9f132d1e101af503f74500d338d3 in conjunction with the use of the Kotlin introspection API. The optimization does not know that the code accesses the members of class `TourbookEntryVerbose`. Hence, it does some internal optimization that breaks the introspection.

Additionally, the issue could not have been seen in a normal instrumentation test because the `R8` optimization is not enabled there. Also, the export works, if the user builds the app in debug mode, which is the default for the development phase, because there the optimization also not enabled.  

**Possible solutions** 

1. Add a rule to file  `app/proguard-rules.pro` for that class to keep the members untouched.
2. Do not use the Kotlin reflection API and accept a minimal boiler-plate code but make it easier to understand and to maintain.

chrgernoe:
TourbookEntryVerbose used Kotlin reflection (memberProperties) to enumerate its fields for export. R8 has no idea reflection reads these fields by name, so minification strips/renames them regardless, silently reducing every exported row to nothing without throwing - verified by inspecting the R8 mapping and DEX output of the release build. Plain JSON export was unaffected since it doesn't use this class or reflection.

Replace the reflective field lookup with an explicit, hand-written list of names/values, and drop the now-unused kotlin-reflect dependency.


app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookEntryVerbose.kt
```diff
@@ -22,30 +22,36 @@ import com.yacgroup.yacguide.database.DatabaseWrapper
...
-    var rockFirstName: String
-    var rockSecondName: String
-    var routeFirstName: String
-    var routeSecondName: String
-    var routeGrade: String
-    var notes: String
-    var date: String
-    var style: String
-    var partners: String
+    val country: String
```
paetz: Can't those be private?

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookEntryVerbose.kt
```diff
@@ -74,16 +80,11 @@ class TourbookEntryVerbose(ascend: Ascend, db: DatabaseWrapper) {
...
-        val self = this
-        return buildMap {
-            TourbookEntryVerbose::class.memberProperties.forEach {
-                if (it.name in keys()) {
-                    put(it.name, it.get(self).toString())
-                }
-            }
-        }
-    }
+    fun asMap(): Map<String, String> = keys().zip(values()).toMap()
```
paetz: A unit test would be great.

paetz (dismissed):

app/src/main/java/com/yacgroup/yacguide/database/tourbook/TourbookEntryVerbose.kt
chrgernoe: There is already a.draft PR for that. See PR #514.

